PR #62, opened 2019-07-01
Use flexbox correctly and just use divs for Media

mdespuits:
To make sure we can configure flexbox properties correctly, we are dropping the custom tag support for `Media` components (for now).

Also, include the `flexbox` utility correctly.


andrewBalekha:
Lgtm

mdespuits:
@andrewBalekha Great! Can I get an approval?

src/components/Media/test.js
```diff
@@ -1,11 +1,11 @@
 import React from 'react';
-import { shallow, mount } from 'enzyme';
+import { mount } from 'enzyme';
 
 import Media from './';
 
 describe('Media', () => {
   describe('Media.Item', () => {
-    it('supports custom tags', () => {
+    xit('supports custom tags', () => {
```
pixelbandito: Asked @mdespuits what this did: `xit`  is just an `ignore` version of `it`, since we intend to make custom tags work later.

PR #63, opened 2019-07-01
Use flexbox correctly and just use divs for Media

mdespuits:
> Duplicate of #62 after master branch cleanup

To make sure we can configure flexbox properties correctly, we are dropping the custom tag support for Media components (for now).

Also, include the `flexbox` utility correctly.


mdespuits:
Accidental re-open

PR #66, opened 2019-07-10
Disable Global CSS

mdespuits:
Currently, if we include `rover-ui` into a project, the CSS custom properties are not being imported correctly and the global styles are messing with font-colors. All of the custom components actually use styled-system now, so for now we don't even need explicit CSS.

This is just a temporary fix to get `rover-ui` into the TK front end application without directly causing unwanted side-effects and so we can start to use the very useful `<Media>` and `<Responsive>` components we've built so far.

### DIFF
--- a/src/components/Button/Addon/style.css
+++ b/src/components/Button/Addon/style.css
@@ -18,7 +18,7 @@
   margin-right: var(--space-lg);
 }
 
-.Addon:first-child {
+.Addon:first-of-type {
   margin-left: 0;
 }
 
--- a/src/index.js
+++ b/src/index.js
@@ -1,8 +1,8 @@
 // Global CSS variables
-import './shared/colors.css';
-import './shared/sizing.css';
-import './shared/variables.css';
-import './shared/reset.css';
+// import './shared/colors.css';
+// import './shared/sizing.css';
+// import './shared/variables.css';
+// import './shared/reset.css';
 
 export { default as Badge } from './components/Badge';
 export { default as Bar } from './components/Bar';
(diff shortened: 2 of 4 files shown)


src/index.js
```diff
@@ -1,8 +1,8 @@
 // Global CSS variables
-import './shared/colors.css';
-import './shared/sizing.css';
-import './shared/variables.css';
-import './shared/reset.css';
+// import './shared/colors.css';
```
pixelbandito: I'm almost certain this will break other imports. We should build a test case on a bare react app that imports all our components, and run using `yarn link` locally to verify.

pixelbandito: I say this ^ because I moved the css import to here from a css module file after discovering that buttons (and I think another component as well) weren't taking any styles when imported into another project.

mdespuits: Ah, you're right. I'd forgotten about the buttons. I'll just remove the reset for now since that's the main culprit

pixelbandito: After my change, you'll need to fix small conflicts around the reset.css. 

Otherwise: 🌻🥪

src/components/Button/Addon/style.css
```diff
@@ -18,7 +18,7 @@
   margin-right: var(--space-lg);
 }
 
-.Addon:first-child {
+.Addon:first-of-type {
```
pixelbandito: Not sure why, but this should be fine. Just curious: why?

mdespuits:
Closing because this isn't quite the right solution and it's already been handled in other changes

PR #67, opened 2019-07-11
feat: add Pill component

andrewBalekha:
I have added a new Pill component that we are going to use in Impact and Influencers pages.

I didn't figure out how svgr works, so I decided to create a react component for an icon.

### DIFF
--- a/src/components/Pill/story.js
+++ b/src/components/Pill/story.js
@@ -7,7 +7,7 @@ import Readme from './README.md';
 function handleClick(id) {
   alert(`clicked ${id}`);
 }
-storiesOf('Uncategorized/Pill', module)
+storiesOf('Star System/Pill', module)
   .addParameters({
     readme: {
       sidebar: Readme,
@@ -16,11 +16,25 @@ storiesOf('Uncategorized/Pill', module)
   .add(
     'Overview',
     () => (
+      <Pill id="pill-1" onClick={handleClick} selected>
+        Pill 1
+      </Pill>
+    ),
+    {
+      info: {
+        inline: true,
+        source: true,
+      },
+    }
+  )
+  .add(
+    'Pills list',
+    () => (
       <React.Fragment>
-        <Pill id="pill-1" onClick={handleClick}>
+        <Pill id="pill-1" onClick={handleClick} selected>
           Pill 1
         </Pill>
-        <Pill selected id="pill-2" onClick={handleClick}>
+        <Pill id="pill-2" onClick={handleClick} selected>
           Pill 2
         </Pill>
         <Pill id="pill-3" onClick={handleClick}>
--- a/src/components/Pill/story.js
+++ b/src/components/Pill/story.js
@@ -1,0 +1,37 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import Pill from './';
+import Readme from './README.md';
+
+function handleClick(id) {
+  alert(`clicked ${id}`);
+}
+storiesOf('Uncategorized/Pill', module)
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+  })
+  .add(
+    'Overview',
+    () => (
+      <React.Fragment>
+        <Pill id="pill-1" onClick={handleClick}>
+          Pill 1
+        </Pill>
+        <Pill selected id="pill-2" onClick={handleClick}>
+          Pill 2
+        </Pill>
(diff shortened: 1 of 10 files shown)


src/icons/clear.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+function ClearIcon(props) {
```
mdespuits: Generally we use `const ClearIcon = (props) => ()` style even for functional components like this one.

src/components/Pill/story.js
```diff
@@ -0,0 +1,37 @@
...
+function handleClick(id) {
+  alert(`clicked ${id}`);
+}
+storiesOf('Uncategorized/Pill', module)
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+  })
+  .add(
```
mdespuits: The typical pattern for these stories uses `Overview` as a place to use Storybook knobs to manipulate the component in whatever was possible to see various possibilities. What's defined here is a bit more what we've gone with `Examples`.

src/components/Pill/story.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+
+import Pill from './';
+import Readme from './README.md';
+
+function handleClick(id) {
+  alert(`clicked ${id}`);
+}
+storiesOf('Uncategorized/Pill', module)
```
mdespuits: This needs to be categorized. I'd put this under `Star System` since it incorporates a separate component and it's completely isolated

src/components/Pill/style.css
```diff
@@ -0,0 +1,51 @@
+@import '../../shared/colors.css';
+
+.pill {
```
mdespuits: I love CSS modules, but we've been leaning more heavily on `styled-system` and `styled-components` for new components. I may be outvoted here, but we should decide on using CSS modules or styled-components exclusively rather than mixing the two.

@pixelbandito @chelshaw Any thoughts?

src/icons/clear.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
```
mdespuits: I'm totally fine with this for now, but we definitely need to figure out how we want to handle icons before we add more after this.

src/components/Pill/story.js
andrewBalekha: I don't know the naming logic, sorry :)

src/components/Pill/style.css
andrewBalekha: css modules are much easier tho. Haven't worked with styled-components.

mdespuits: @andrewBalekha I definitely agree. In my experience with `styled-components` it has already proven to be tricky. We were hoping to take advantage of some features that `styled-system` and `styled-components` provided but it may be more overhead than it's worth.

@pixelbandito may have some more thoughts on this after a conversation I had with him yesterday afternoon. Anything to weigh in here?

pixelbandito: I'm trying to make styled components work with hover states in another branch. My take right now is that it's way too hard. 

I'm ok with going back to plain old css, but we should do a pros/cons in Slack or a GitHub issue and give other contributors and front end folks a chance to weigh in.

andrewBalekha:
I will move it to `tk-frontend-app` because it is blocking us.

andrewBalekha:
> Just a note: in the near future we will want other actions besides clear (namely, a kebab with dropdown menu) but seems like we should be able to update with that functionality simply enough.

I would try to keep components as simple as they can be. And provide render prop to render action with default clear action.

src/components/Pill/style.css
```diff
@@ -0,0 +1,51 @@
+@import '../../shared/colors.css';
```
pixelbandito: PR #69 should have already made colors.css available here, so you can remove this line.

src/components/Pill/style.css
pixelbandito: After polling, and talking at the meeting, I think we should table the discussion and allow CSS modules to proliferate, so @andrewBalekha's code here is mostly fine.

However, we've been consistently PascalCasing CSS module class names when they refer to the main wrapper of a component and share the component's name. 

So it should be `.Pill {`

pixelbandito:
I rebased off this, updated it, and got it into master. Closing this PR.

PR #69, opened 2019-07-16
Add babel and postcss plugins/config for IE 11, prefix css custom prop names

pixelbandito:
* Changed example app's link to point to built assets (you have to run `yarn build` in the project root first)
* Added minimal examples to example/App.js so I could test IE 11 compatibility
* Linter fixes on example/index.js
* Added plugins to support css `calc()` and js `Object.values()`
* Configured postcss custom property plugin to expose css variables from the shared folder
* Prefixed all our CSS custom properties with `rvr-` (RoVeR). _This isn't necessary with the current config, since we're not preserving the custom props, but we've talked about exposing them in the future and it seemed sensible_.
* Removed all references to `reset.css`, since using it was breaking apps that imported our library.
* Moved `src/components/Button/colors.css` to `shared/buttons.css` so rollup config doesn't have to enumerate a bunch of component-specific paths. We should only be creating css custom prop files for pretty core theme stuff, so I guess this is better.


src/shared/variables.css
```diff
@@ -1,9 +1,11 @@
...
   /* Colors */
-  --box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.2);
+  --rvr-box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.2);
 
   /* Fonts */
-  --base-font-color: #58666e;
-  --base-font-family: 'Source Sans Pro', sans-serif;
-  --fontFamily-mono: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
+  --rvr-base-font-color: #58666e;
+  --rvr-base-font-family: 'Source Sans Pro', sans-serif;
```
mdespuits: Should we go ahead and switch this to the Proxima font while we are updating this?

pixelbandito: Yes, but no. All the colors have to get updated too, I just don't want to mix concerns with this PR any more than they already are.

PR #71, opened 2019-07-17
Component side tray

chelshaw:
**SUMMARY**
Add a sidetray component that calls closeCallback when:
- You click outside the sidetray
- You click the `esc` key

Has props for header and footer nodes, and content between them automatically flexGrows


src/components/SideTray/index.js
```diff
@@ -0,0 +1,68 @@
...
+        </div>
+      </div>
+      {clickOffBackdrop}
+    </div>
+  );
+};
+
+SideTray.defaultProps = {
+  visible: false,
+  modifiers: [],
```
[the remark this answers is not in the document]
chelshaw: oh yeah. I removed it from props but not defaults :D

src/components/SideTray/style.css
```diff
@@ -0,0 +1,56 @@
...
+.main {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-evenly;
+  overflow: hidden;
+  height: 100%;
+}
+
+.content {
+  flex-grow: 1;
```
pixelbandito: Any time something resizes with flex, you need `min-width: 0` / `min-height: 0` (depending on the flow).

src/components/SideTray/style.css
```diff
@@ -0,0 +1,56 @@
...
+  right: 0;
+  bottom: 0;
+  left: 0;
+  height: 100vh;
+  width: 100vw;
+  z-index: var(--rvr-off-canvas-drawer-z-index) - 1;
+  background-color: rgba(0, 0, 0, 0.3);
+  overflow: auto;
+  opacity: 10;
+  transition: opacity 150ms var(--rvr-easeInOutQuad);
```
pixelbandito: The variables for timing functions are a good start. Wonder if we could come up with a generic pattern to inform when to use what? (I just posted about [this](https://uxdesign.cc/the-ultimate-guide-to-proper-use-of-animation-in-ux-10bd98614fa9) in eng-guild-front-end)

I was thinking something like

Timing function variables
* timing-fade
* timing-move-in
* timing-move-out

Durations
* slow
* medium
* fast

For this element, since it's pretty big, we can do "slow timing-fade" for the backdrop and "slow timing-move-in"/"slow timing-move-out" for the actual tray. We'd want to go back and use a "fast" duration for buttons, and we could even come back later and adjust speeds based on media queries.

src/components/SideTray/README.md
```diff
@@ -0,0 +1,3 @@
+# <SideTray>
+
+Add your documentation for the `SideTray` component here.
```
pixelbandito: This would be nice.

example/yarn.lock
```diff
@@ -217,9 +217,9 @@ ajv@^5.0.0, ajv@^5.1.5, ajv@^5.2.0:
     json-schema-traverse "^0.3.0"
 
 ajv@^6.0.1, ajv@^6.5.5:
-  version "6.10.1"
-  resolved "https://trendkite.jfrog.io/trendkite/api/npm/libs-frontend/ajv/-/ajv-6.10.1.tgz#ebf8d3af22552df9dd049bfbe50cc2390e823593"
-  integrity sha1-6/jTryJVLfndBJv75QzCOQ6CNZM=
+  version "6.10.2"
```
pixelbandito: Do you have time to add a .gitattributes file for this?
```
yarn.lock             -diff
example/yarn.lock     -diff
```

src/components/SideTray/story.js
```diff
@@ -0,0 +1,46 @@
...
+  );
+}
+
+storiesOf('Star System/SideTray', module)
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+  })
+  .add('Overview', () => <SideTrayTest />, {
```
pixelbandito: This is the "Example"
The "Overview" should be the bare component with only knobs and actions

pixelbandito:
I committed a change to add a SideTray to the example page. Would you verify it doesn't break your subsequent changes?

Also, I tested in IE11 and Edge - looking good!!

src/components/SideTray/style.css
chelshaw: I'm not gonna do this at the moment; i'm not even convinced 150ms is a quantity worth transitioning over

src/components/SideTray/Body/index.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import style from './style.css';
+
+const Body = ({ children, ...passedProps }) => {
+  return (
+    <div className={style.SideTray__Body} {...passedProps}>
```
pixelbandito: Style name could just be `Body`, since css modules should be taking care of uniqueness.

And you should probably concat `style.Body` to whatever `props.className` the consumer passed in.

src/components/SideTray/Body/index.js
```diff
@@ -0,0 +1,17 @@
...
+      {children}
```
pixelbandito: You don't need to enumerate children. It's just a prop, so let `passedProps` handle it.

src/components/SideTray/Footer/index.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import style from './style.css';
+
+const Footer = ({ children, ...passedProps }) => {
+  return (
+    <div className={style.SideTray__Footer} {...passedProps}>
```
pixelbandito: Same comments as body.

src/components/SideTray/Header/index.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import style from './style.css';
+
+const Header = ({ children, ...passedProps }) => {
```
pixelbandito: 🔁

src/stories.js
```diff
@@ -1,5 +1,6 @@
 import './shared/colors.css';
 import './shared/sizing.css';
+import './shared/buttons.css';
```
pixelbandito: I'm not 100%, but we might be able to drop these since we're doing the CSS loader config to make the variables available.

chelshaw: when this was missing my buttons in storybook weren't working

src/components/SideTray/Body/index.js
mdespuits: If you take a look at the `Media.jsx` in tk-ui I have a pattern that does this exact thing if that helps

src/stories.js
mdespuits: This was not working in Storybook because the loader changes were made in the rollup config, afaik

src/components/SideTray/style.css
```diff
@@ -1,5 +1,5 @@
-.sideTray {
-  width: var(--rvr-side-tray-width);
```
mdespuits: Why are we losing the variable here?

pixelbandito:
Good catch on the imports in stories.i think that's because storybook used a different webpack config and therefore different css loader module options, 

Importing shared css directly for now in stories is great.

src/components/SideTray/style.css
pixelbandito: Hey @mdespuits,

I suggested this. I figured that:
1. The sidetray width isn't intended for re-use among components. 
2. There's nothing magical about 400 px in our sizing system.
3. It's not part of the theme.

I could see us standardizing on some sizes for modals / popovers / etc. in the future, but for now I thought this was cleaner than having a block of sidebar-specific properties in our global variables.

src/stories.js
pixelbandito: Yup, so this is good.

pixelbandito:
I'm drunk with power and seeing if I can dismiss @mdespuits review, but only because I already verified that @chelshaw made the changes requested.

PR #72, opened 2019-07-22
Add SideTray direction, width, height, plus pass through style and className props

pixelbandito:


src/components/SideTray/index.js
```diff
@@ -25,35 +40,84 @@ const SideTray = ({ visible, onClose, children }) => {
...
+  onClose: PropTypes.func.isRequired,
+  style: PropTypes.object,
+  visible: PropTypes.bool,
+  width: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+};
+
+SideTray.defaultProps = {
+  className: '',
+  direction: 'r',
+  footer: null,
```
chelshaw: footer and header props can be removed

PR #74, opened 2019-07-26
Add TabMenu component, including Item and SimpleTabMenu

chelshaw:
**SUMMARY**
Add tabMenu (which exists in TK-UI and APP buuuuuut I wanted it to be more usable out of the box. 

sooooo as part of this, we have a SimpleTabMenu (which just takes an array and a couple other props) and a rather simple TabMenu + Item set of components which can be customized for behavior but manage to keep the same look as all the other tab menus.

This will be used extensively in iHub
Screen Shot 2019-07-25 at 10 08 54 PM

### DIFF
--- a/src/components/TabMenu/Item/index.js
+++ b/src/components/TabMenu/Item/index.js
@@ -1,0 +1,48 @@
+import React from 'react';
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+
+import style from './style.css';
+
+const Item = ({
+  size,
+  className: customClassName,
+  active,
+  onClick,
+  children,
+  ...props
+}) => {
+  const isChildActive = child => {
+    if (typeof child !== 'object') return false;
+    if ((child.props || {}).className) {
+      return child.props.className.indexOf('active') >= 0;
+    }
+    return false;
+  };
+
+  const className = classNames(style.TabItem, customClassName, {
+    [style.active]: active || isChildActive(children),
+  });
+  return (
+    <li className={className} {...props}>
+      {children}
+    </li>
+  );
+};
+
+Item.propTypes = {
+  size: PropTypes.number,
+  className: PropTypes.string,
+  active: PropTypes.bool,
+  onClick: PropTypes.func,
+  children: PropTypes.node.isRequired,
+};
+
+Item.defaultProps = {
+  size: 16,
+  className: '',
+  active: false,
+  onClick: () => {},
+};
+
+export default Item;
--- a/src/components/TabMenu/Item/index.js
+++ b/src/components/TabMenu/Item/index.js
@@ -4,23 +4,15 @@ import PropTypes from 'prop-types';
 
 import style from './style.css';
 
-const Item = ({
-  size,
-  className: customClassName,
-  active,
-  onClick,
-  children,
-  ...props
-}) => {
+const Item = ({ className: customClassName, active, children, ...props }) => {
   const isChildActive = child => {
-    if (typeof child !== 'object') return false;
-    if ((child.props || {}).className) {
+    if (React.isValidElement(child)) {
       return child.props.className.indexOf('active') >= 0;
     }
     return false;
   };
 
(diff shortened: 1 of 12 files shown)


src/components/TabMenu/index.js
```diff
@@ -0,0 +1,61 @@
...
+import classNames from 'classnames';
+
+import Item from './Item';
+import style from './style.css';
+
+const TabMenu = ({ align, className, ...props }) => {
+  const tabMenuClass = classNames(
+    style.TabMenu,
+    className,
+    style[`${align}Align`]
```
pixelbandito: Super minor: `styled-system` has a utility for flex styles, if we still feel ok about that pattern. One `flex` util in classNames would let the consumer apply a whole host of flex styles.

Even if we don't double down on styled-system, what would you think of using `justifyContent` instead of `align` as the prop type? It will be more transparent to the end user what they're changing.

src/components/TabMenu/index.js
```diff
@@ -0,0 +1,61 @@
...
+  );
+  return <ul className={tabMenuClass} {...props} />;
+};
+
+TabMenu.Item = Item;
+export const itemPadding = style.itemPadding;
+
+TabMenu.propTypes = {
+  align: PropTypes.oneOf(['left', 'center', 'right']),
```
pixelbandito: `flex-start`/`flex-end`?

src/components/TabMenu/Item/index.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+    return false;
+  };
+
+  const className = classNames(style.TabItem, customClassName, {
+    [style.active]: active || isChildActive(children),
+  });
+  return (
+    <li className={className} {...props}>
+      {children}
```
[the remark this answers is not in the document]
chelshaw: Oh right. This was from when i was experimenting with modifying the child. Updating!

src/components/TabMenu/Item/style.css
```diff
@@ -0,0 +1,20 @@
+.TabItem {
+  color: var(--rvr-gray-lite-1);
+  font-weight: normal;
+  padding: 0;
+  margin: 0 12px;
+  display: block;
+  border-bottom: 2px solid transparent;
+  box-sizing: content-box;
```
mdespuits: Why `content-box`? Seems unusual to explicitly have this enabled?

src/components/TabMenu/style.css
```diff
@@ -0,0 +1,37 @@
...
+  /* reset default button styles */
+  background-color: transparent;
+  border: none;
+  color: inherit;
+  text-decoration: none;
+  display: block;
+  border-radius: 0;
+}
+
+.xsTextSize { font-size: 14px; }
```
mdespuits: Are there no font-sizes as css custom properties in Rover yet? Either way, this should be configurable rather than static

src/components/TabMenu/index.js
chelshaw: I don't love it, simply because at that point you might as well put all this stuff in a passed style prop. The point of SimpleTabMenu is to be as simple and human-readable as possible... but i can be swayed i guess

src/components/TabMenu/Item/index.js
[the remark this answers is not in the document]
chelshaw: Oh, i had to pull out children to check if it's active above. Leaving as-is

src/components/TabMenu/Item/style.css
chelshaw: So that the bottom border doesn't rest above any parent border styles

src/components/TabMenu/style.css
chelshaw: Good catch, i put these as placeholder because we didn't have css variables for font sizes other than small yet... but this is a good reminder to standardize on what those sizes should be

mdespuits:
Sounds great! Thanks!

PR #77, opened 2019-07-30
0.1.0-alpha.11 // Update tab menu

chelshaw:
light style updates to TabMenu.Item, and export `EasyTabMenu` so we can switch over to that in the app and remove `SimpleTabMenu`

### DIFF
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -6658,15 +6658,9 @@ react-scripts@^1.1.4:
   optionalDependencies:
     fsevents "^1.1.3"
 
-react@^16.8.0:
-  version "16.8.6"
-  resolved "https://trendkite.jfrog.io/trendkite/api/npm/libs-frontend/react/-/react-16.8.6.tgz#ad6c3a9614fd3a4e9ef51117f54d888da01f2bbe"
-  integrity sha1-rWw6lhT9Ok6e9REX9U2IjaAfK74=
-  dependencies:
-    loose-envify "^1.1.0"
-    object-assign "^4.1.1"
-    prop-types "^15.6.2"
-    scheduler "^0.13.6"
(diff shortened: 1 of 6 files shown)


example/yarn.lock
```diff
@@ -6658,15 +6658,9 @@ react-scripts@^1.1.4:
...
-react@^16.8.0:
-  version "16.8.6"
-  resolved "https://trendkite.jfrog.io/trendkite/api/npm/libs-frontend/react/-/react-16.8.6.tgz#ad6c3a9614fd3a4e9ef51117f54d888da01f2bbe"
-  integrity sha1-rWw6lhT9Ok6e9REX9U2IjaAfK74=
-  dependencies:
-    loose-envify "^1.1.0"
-    object-assign "^4.1.1"
-    prop-types "^15.6.2"
-    scheduler "^0.13.6"
+"react@link:../node_modules/react":
```
pixelbandito: This should get removed before commit, right?

src/components/TabMenu/Item/style.css
```diff
@@ -8,6 +8,14 @@
   box-sizing: content-box;
 }
 
+.Item:first-of-type {
```
pixelbandito: I know the linter warns about this, but it's not exactly the same thing.
If you need to support ...
```jsx
<span /><Item />
```
... and Item renders a div, it will break; because Item is still the first of its type

May not be an issue for tabs, but worth checking.

chelshaw: I'm not totally worried about any other divs, because if the behavior is consistent then the first <span /> or whatever can apply its own spacing. Is that a reasonable call to make?

pixelbandito: Works for me.

PR #78, opened 2019-07-30
Add hygen templates for example app

pixelbandito:
No-op as far as the API, so I'm not bumping the version here.


example/src/App.js
```diff
@@ -137,6 +137,7 @@ const App = () => {
           />
         </div>
       </section>
+      {/** USAGE_INJECTOR */}
```
chelshaw: 😍

PR #81, opened 2019-07-31
Add Pill, DeletablePill, EasyPill

pixelbandito:
I picked up @andrewBalekha's changes from a couple of weeks ago, and split the `Pill` so there's a very un-opinionated version in addition to one with a built-in delete action.

I also started to stub out an `EasyPill`, but it's not ready for use - the actions and callbacks work, but they need to be put into a dropdown, which doesn't exist in RoverUI yet.

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -11,6 +11,10 @@ import {
   Responsive,
   SideTray,
   EasyTabMenu,
+  Pill,
+  DeletablePill,
+  EasyPill,
+  // IMPORT_INJECTOR
 } from '@cision/rover-ui';
 
 const App = () => {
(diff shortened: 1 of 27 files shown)


example/src/App.js
```diff
@@ -11,6 +11,10 @@ import {
   Responsive,
   SideTray,
   EasyTabMenu,
+  Pill,
+  DeletablePill,
+  EasyPill,
+  // IMPORT_INJECTOR
```
pixelbandito: Missed this in my rebase! Needs to be there for hygen to work it's magic

src/components/TabMenu/index.js
```diff
@@ -26,11 +26,7 @@ TabMenu.defaultProps = {
   className: '',
 };
 
-/*
- * SimpleComponent naming convention is deprecated due to confusion.
- * Use EasyComponent moving forward.
- */
-export const SimpleTabMenu = ({ tabs, activeTab, size = 'sm', ...props }) => {
+export const EasyTabMenu = ({ tabs, activeTab, size = 'sm', ...props }) => {
```
pixelbandito: I changed the primary exported component to be `EasyTabMenu` so it'll be easier to rip out the old one.

Sorry for stepping on your toes, @chelshaw! I wrote most of this last night on the `Pill` branch when adding `EasyPill`, for consistency. Just renamed stuff.

src/shared/colors.css
```diff
@@ -53,4 +53,65 @@
 
   /* Notify */
   --rvr-color-notify: var(--rvr-hot-pink);
+
+  /* colors from zeplin styleguide, created by Egle */
+  /* stylelint-disable comment-empty-line-before */
+
+  /* --teal: #00837e; */
```
pixelbandito: This is from @andrewBalekha's initial pill commit, which I rebased with my changes.

I know we need to re-map the main color variables so we're not hosting both pre and post-rebranding colors.
But, since I don't have the mapping handy, I thought I'd leave the data in here until next time.

src/components/TabMenu/index.js
chelshaw: I thought... i already did this in the previously merged code? but good call on switching the default

src/components/DeletablePill/README.md
```diff
@@ -0,0 +1,6 @@
+# DeletablePill
+
+When checked, clicking anywhere on a deletable pill fires an `onDelete` function.
```
chelshaw: Not sure the deletable pill should always assume clicking on it deletes it. Especially if there's also the clear icon. Also, i find naming the x icon "clear" is sooooo confusing... i thought it was just a blank space like we have in tk-ui for the longest time

PR #82, opened 2019-08-05
Sidetray style fix

chelshaw:
Accidentally had the shrink set to 1, where it should have been 0. Caused some issues with implementation, but this fixes it.

### DIFF
--- a/src/components/SideTray/style.css
+++ b/src/components/SideTray/style.css
@@ -46,13 +46,13 @@
 .header {
   padding: 12px;
   border-bottom: 1px solid var(--rvr-gray-lite-5);
-  flex: 0 1 auto;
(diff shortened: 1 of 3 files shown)


src/components/SideTray/style.css
```diff
@@ -46,13 +46,13 @@
 .header {
```
mdespuits: _non-blocking_

Since this is actually applying to a sub-component itself, I think this should be capitalized or treated like a top-level CSS module class.

mdespuits:
@chelshaw I think that we should leave version bumps to separate pull requests so that there are no potential conflicts between branches if multiple pull requests bump the version

pixelbandito:
👍 good stuff

PR #83, opened 2019-08-05
Media uses CSS modules instead of styled-system

mdespuits:
This way, CSS classes from the main application can override any default styles since styled-components injects CSS later in the load process.


src/components/Media/index.js
```diff
@@ -1,13 +1,41 @@
...
+import { space, flex } from 'styled-system';
 
-const Media = StyledMedia;
-Media.displayName = 'Media';
+import styles from './style.css';
+import withDefaultTheme from '../withDefaultTheme';
 
-const Body = StyledBody;
-Body.displayName = 'Media.Body';
+const FlexSpacer = withDefaultTheme(styled.div(flex, space));
```
chelshaw: Is this going to have similar issues to what we were seeing before?

src/components/Media/style.css
```diff
@@ -0,0 +1,21 @@
...
+  grid-template-columns: 100%;
+}
+
+.Body {
+  flex: 1 1 auto;
+  min-width: 0;
+}
+
+.Item {
+  flex: 0 0 auto;
```
chelshaw: we need min-width here too, right?

src/components/Media/index.js
mdespuits: No, this only provides the component props that would allow us to _set_ those values when we want. These don't add any styles by default

src/components/Media/style.css
mdespuits: Ah, possibly. Can't hurt to be safe

PR #89, opened 2019-08-08
Add EasyPill

pixelbandito:
Adding the pill component optimized for ease-of-use instead of extensibility.

Pass in a child, and optionally an actions array and/or onDelete callback, and it handles the rest.

### DIFF
--- a/src/components/Media/index.js
+++ b/src/components/Media/index.js
@@ -2,12 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import styled from 'styled-components';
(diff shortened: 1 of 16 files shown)


src/components/Media/index.js
```diff
@@ -2,12 +2,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 import styled from 'styled-components';
-import { space, flex } from 'styled-system';
+import { alignSelf, space, flex } from 'styled-system';
```
mdespuits: Isn't `alignSelf` part of the `flex`?

PR #94, opened 2019-08-19
EVER-1474 modify sidetray to allow outside click and margin top for header to be visible

kcj002:
Screen Shot 2019-08-19 at 11 57 17 AM
Screen Shot 2019-08-19 at 11 57 14 AM

Changes: adds new arguments for the SideTray to allow for marginTop & have no backdrop/close on outside click. These new requirements for the sidetray will be used in a handfull of other areas throughout the application.

Risk: low, small isolated changes to SideTray

### DIFF
--- a/src/components/SideTray/index.js
+++ b/src/components/SideTray/index.js
@@ -49,7 +49,7 @@ const SideTray = ({
       <button className={style.backdrop} onClick={onClose} />
     ) : null;
 
-  // Handle custom widths / heights / directions / top margin
+  // Handle custom widths / heights / directions
   const parsedHeight = parseCssSize({ size: height });
   const parsedWidth = parseCssSize({ size: width });
   let hideTransformStyle;
--- a/src/components/SideTray/index.js
+++ b/src/components/SideTray/index.js
@@ -13,10 +13,10 @@ const SideTray = ({
   closeOnOutsideClick,
   direction,
   height,
-  marginTop,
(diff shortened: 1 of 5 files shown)


src/components/SideTray/index.js
```diff
@@ -10,8 +10,10 @@ import style from './style.css';
 const SideTray = ({
   children,
   className,
+  closeOnOutsideClick,
   direction,
   height,
+  marginTop,
```
pixelbandito: I see a couple of issues with this approach.
1. `marginTop` implies that there should be configurability on other margins, and, since it's just a CSS style override, I'd rather see us spread a `styles` prop here (or something slightly more specific like `wrapperStyles`/`backdropStyles`).
2. Since we're setting a height to `100vh` by default (for both backdrop and content, I think), the margin will technically be pushing content off the bottom of the page.

I think, ideally, we'd put both backdrop and side tray content in a single wrapper div, add a `wrapperStyles` there, that applies `height: 100vh; width: 100vw; box-sizing: border-box; position: fixed;` by default. We'd have side tray and backdrop use relative positioning and `height: 100%;` as their default. 

Then, for this use case, set the `wrapperStyle` to `paddingTop: 40px;` and you get the child sizing for free.

kcj002: Thank you @pixelbandito ! Super helpful

PR #95, opened 2019-08-20
Add social icons with default colors

chelshaw:
Adds social logo icons with default colors, and ability to override if passed "fill" prop with proper hex or color


storybook-static/iframe.html
```diff
@@ -0,0 +1,61 @@
+<!doctype html><html><head><meta charset="utf-8"><title>Storybook</title><meta name="viewport" content="width=device-width,initial-scale=1"><base target="_parent"><style>:not(.sb-show-main) > .sb-main,
+  :not(.sb-show-nopreview) > .sb-nopreview,
```
kcj002: Question because I'm new to this project and I opted to not include these in my PR (perhaps incorrectly): should this storybook-static directory be included in pull requests?

pixelbandito: Probably not.

src/components/Icon/icons/colors.js
```diff
@@ -0,0 +1,12 @@
+const iconColorsMap = {
+  facebook: '#3B579D',
+  googleplus: '#DB4437',
+  instagram: '#0A0B09',
+  linkedin: '#0077B5',
+  reddit: '#FF4301',
+  twitch: '#6441A4',
+  twitter: '#1DA1F2',
+  youtube: 'red',
```
pixelbandito: Is youtube really just 'red'? If so, awesome.

storybook-static/iframe.html
chelshaw: derp, forgot to remove these. If we're all in agreement I will also add it to the gitignore

src/components/Icon/icons/colors.js
chelshaw: according to our current icon!

.gitignore
```diff
@@ -23,3 +23,4 @@ coverage
 npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
+storybook-static
```
kcj002: 👏

PR #96, opened 2019-08-21
Added note about npm accounts to publication instructions

pixelbandito:


pixelbandito:
This is stupid and simple, but I realized when walking Kate through that we don't mention you need the npms.

PR #101, opened 2019-09-03
Collapse, ExpansionPanel and Accordion components

hpieroni:
This pr includes the following components:
- `Chevron` icon
- `Collapse`: wraps `Transition` component from [react-transition-group](http://reactcommunity.org/react-transition-group/transition) and animates height.
- `ExpansionPanel`: includes a header (with an optional icon) and a body that is collapsed/expanded via `Collapse` component. It could be a controlled or uncontrolled component.
- `Accordion`: receives multiple `ExpansionPanel` components as children and allows only one of them to be expanded at the same time.

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -17,6 +17,7 @@ import {
   Dropdown,
   EasyDropdown,
   Collapse,
+  ExpansionPanel,
   // IMPORT_INJECTOR
 } from '@cision/rover-ui';
 
@@ -284,6 +285,13 @@ const App = () => {
       <section>
         <h1>Collapse</h1>
         <Collapse isOpen={true}>Collapse content</Collapse>
+      </section>
+      <section>
+        <h1>ExpansionPanel</h1>
+        <ExpansionPanel defaultExpanded={false}>
+          <ExpansionPanel.Header>Click me!</ExpansionPanel.Header>
+          <ExpansionPanel.Body>├── Body</ExpansionPanel.Body>
+        </ExpansionPanel>
       </section>
       {/** USAGE_INJECTOR */}
     </div>
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -284,7 +284,7 @@ const App = () => {
       </section>
       <section>
         <h1>Collapse</h1>
-        <Collapse isOpen={true}>Collapse content</Collapse>
+        <Collapse isOpen>Collapse content</Collapse>
       </section>
       <section>
         <h1>ExpansionPanel</h1>
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -16,6 +16,7 @@ import {
   EasyPill,
   Dropdown,
   EasyDropdown,
+  Collapse,
   // IMPORT_INJECTOR
 } from '@cision/rover-ui';
 
@@ -279,6 +280,10 @@ const App = () => {
         >
           With actions and onDelete
         </EasyPill>
+      </section>
+      <section>
+        <h1>Collapse</h1>
+        <Collapse isOpen={true}>Collapse content</Collapse>
       </section>
       {/** USAGE_INJECTOR */}
     </div>
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -293,6 +293,19 @@ const App = () => {
           <ExpansionPanel.Body>├── Body</ExpansionPanel.Body>
         </ExpansionPanel>
       </section>
+      <section>
+        <h1>Accordion</h1>
+        <Accordion>
+          <ExpansionPanel>
+            <ExpansionPanel.Header>Header 1</ExpansionPanel.Header>
+            <ExpansionPanel.Body>Body 1</ExpansionPanel.Body>
+          </ExpansionPanel>
+          <ExpansionPanel>
+            <ExpansionPanel.Header>Header 2</ExpansionPanel.Header>
+            <ExpansionPanel.Body>Body 2</ExpansionPanel.Body>
+          </ExpansionPanel>
+        </Accordion>
+      </section>
       {/** USAGE_INJECTOR */}
     </div>
   );
--- a/src/components/Accordion/index.js
+++ b/src/components/Accordion/index.js
@@ -1,0 +1,47 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+const Accordion = ({ defaultExpandedPanel, children, ...passedProps }) => {
+  const [expandedPanel, setExpandedPanel] = useState(defaultExpandedPanel);
+
+  const handleToggle = (panel, onToggle) => (event, expanded) => {
+    const expandedPanel = expanded ? panel : false;
+    setExpandedPanel(expandedPanel);
+    if (onToggle) {
+      onToggle(event, expanded);
+    }
+  };
+
+  return (
+    <div {...passedProps}>
+      {React.Children.map(children, (child, index) =>
+        React.cloneElement(child, {
+          ...child.props,
+          defaultExpanded: undefined,
+          expanded: index === expandedPanel,
+          onToggle: handleToggle(index, child.props.onToggle),
+        })
+      )}
+    </div>
+  );
+};
+
+Accordion.propTypes = {
+  /**
+   * Group of expansion panels.
+   */
+  children: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.arrayOf(PropTypes.node),
+  ]),
+  /**
+   * Index of default expanded panel starting from 0.
+   */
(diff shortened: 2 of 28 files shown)


src/components/Accordion/index.js
```diff
@@ -0,0 +1,47 @@
+import React, { useState } from 'react';
+import PropTypes from 'prop-types';
+
+const Accordion = ({ defaultExpandedPanel, children, ...passedProps }) => {
+  const [expandedPanel, setExpandedPanel] = useState(defaultExpandedPanel);
+
+  const handleToggle = (panel, onToggle) => (event, expanded) => {
+    const expandedPanel = expanded ? panel : false;
+    setExpandedPanel(expandedPanel);
```
pixelbandito: I don't think you should switch type of `expandedPanel` from a number to `false`. Maybe a number or `null` would be cleaner?

src/components/Accordion/index.js
```diff
@@ -0,0 +1,47 @@
...
+    PropTypes.arrayOf(PropTypes.node),
+  ]),
+  /**
+   * Index of default expanded panel starting from 0.
+   */
+  defaultExpandedPanel: PropTypes.number,
+};
+
+Accordion.defaultProps = {
+  children: null,
```
pixelbandito: If you have your eslinter enabled, it should complain about not having a default value for `defaultExpandedPanel`

src/components/Accordion/story.js
```diff
@@ -0,0 +1,123 @@
...
+import Readme from './README.md';
+
+storiesOf('Galaxies/Accordion', module)
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+  })
+  .add(
+    'Overview',
```
pixelbandito: The "Overview" story should be fully controlled with knobs (https://www.npmjs.com/package/@storybook/addon-knobs) for every prop.

The "Example" is where you can provide state and make interactive demos of various pieces of functionality.

src/components/Accordion/test.js
```diff
@@ -0,0 +1,132 @@
...
+    );
+    wrapper.children().forEach(panel => {
+      expect(panel.prop('expanded')).toBe(false);
+    });
+  });
+
+  it('defaultExpanded` should set a panel a `expanded`', () => {
+    const wrapper = shallow(
+      <Accordion defaultExpandedPanel={1}>
+        <ExpansionPanel>
```
pixelbandito: Since the intent is for `ExpansionPanel` to be inside an `Accordion` every time, you should use the pattern `Accordion.ExpansionPanel` instead. You can see an example with `SideTray`/`SideTray.Footer` in `SideTray/index.js`

src/components/Collapse/index.js
```diff
@@ -0,0 +1,112 @@
...
+  isOpen: PropTypes.bool,
+  /**
+   * The duration for the transition, in milliseconds.
+   */
+  timeout: PropTypes.number,
+};
+
+Collapse.defaultProps = {
+  ...Transition.defaultProps,
+  isOpen: false,
```
pixelbandito: Ditto that the linter should complain if you're not providing a `children` default prop and `children` is not marked required.

src/components/ExpansionPanel/index.js
```diff
@@ -0,0 +1,90 @@
...
+ExpansionPanel.propTypes = {
+  /**
+   * The content of the expansion panel.
+   */
+  children: props => {
+    const children = React.Children.toArray(props.children);
+    if (isEmpty(children)) {
+      return new Error('ExpansionPanel: children is required.');
+    }
+    if (children[0].type !== Header) {
```
pixelbandito: I like this very much, but test it carefully. iirc, there used to be issues where hot reloader or something in the build process would stop this simple type check from working.

The old recommendation was to add a custom static type to the component, like `Header.displayName = "XXX.Header"`, and check that.

However, this way is better, if it works. Maybe this was something that's changed in recent versions.

pixelbandito:
👏 👏 👏 
Nice job with tests, and very thorough! I left a few comments, but this looks like an awesome contribution to RoverUI!

src/components/Accordion/index.js
hpieroni: You're right! I didn't realize that

src/components/Accordion/story.js
hpieroni: @pixelbandito The only prop I can think about is `style`.
`defaultExpandedPanel` is only used to set the initial state of the component. I could watch for its changes using `useEffect` and update the state. What do you think?

src/components/Accordion/test.js
hpieroni: I created the `Accordion` component as a wrapper that allows only an `ExpansionPanel` to be opened at the same time. You definitely are able to use `ExpansionPanel` out of the scope of an `Accordion`

src/components/Accordion/index.js
```diff
@@ -0,0 +1,47 @@
...
+      )}
+    </div>
+  );
+};
+
+Accordion.propTypes = {
+  /**
+   * Group of expansion panels.
+   */
+  children: PropTypes.oneOfType([
```
mdespuits: We have a `propTypes.js` file that we could put this in since this is probably going to be something we could use a few times.

src/components/ExpansionPanel/Body/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+const Body = ({ children, ...passedProps }) => (
+  <div {...passedProps}>{children}</div>
+);
+
+Body.propTypes = {
+  /**
+   * The content to render.
+   */
+  children: PropTypes.oneOfType([
```
mdespuits: Repeat mention of the shared `propTypes.js` file that could share this

src/components/ExpansionPanel/Body/index.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Body = ({ children, ...passedProps }) => (
```
mdespuits: `children` is one of the props you can spread to other components without having to explicitly pass them. If we are going to keep this pass-thru component, you can clean it up to just one line:

```js
const Body = ({ ....props }) => <div {...props} />;
```

src/components/ExpansionPanel/Header/index.js
```diff
@@ -0,0 +1,61 @@
...
+      )}
+    </div>
+  );
+};
+
+Header.propTypes = {
+  /**
+   * The header content to render.
+   */
+  children: PropTypes.oneOfType([
```
mdespuits: Another mention of `propTypes.js` shared type

src/components/ExpansionPanel/Body/index.js
hpieroni: I think that `PropTypes.node` is what I was looking for

src/components/ExpansionPanel/index.js
hpieroni: not sure about this

src/components/Accordion/story.js
pixelbandito: Might be a crazy thought, but you could add a `key` based on the `defaultExpandedPanel` prop, and then React should remount the component of the user changes that one prop.

src/components/ExpansionPanel/index.js
pixelbandito: If it's working, this is fine.

src/components/ExpansionPanel/Body/index.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Body = props => <div {...props} />;
+
+Body.propTypes = {
+  /**
+   * The content to render.
+   */
+  children: PropTypes.node,
```
pixelbandito: Since you're just passing the props, you don't need to add the propTypes anymore.

example/src/App.js
```diff
@@ -280,6 +282,30 @@ const App = () => {
...
+      <section>
+        <h1>ExpansionPanel</h1>
+        <ExpansionPanel defaultExpanded={false}>
+          <ExpansionPanel.Header>Click me!</ExpansionPanel.Header>
+          <ExpansionPanel.Body>├── Body</ExpansionPanel.Body>
+        </ExpansionPanel>
+      </section>
+      <section>
+        <h1>Accordion</h1>
+        <Accordion>
```
pixelbandito: You didn't import `Accordion`. :(

hpieroni: https://github.com/cision/rover-ui/pull/104

PR #102, opened 2019-09-09
Chore/add responsive show hide

pixelbandito:
Adding 2 new responsive components, for simple showing and hiding of components.

### DIFF
--- a/src/components/Responsive/Visible/index.js
+++ b/src/components/Responsive/Visible/index.js
@@ -4,15 +4,10 @@ import intersection from 'lodash/intersection';
 
 import Context from '../Context';
 
-const Visible = ({
-  children,
-  responsiveContext: VisibleResponsiveContext,
-  ...passedProps
-}) => {
+const Visible = ({ children, breakpoints, ...passedProps }) => {
   const getIsVisible = useCallback(
-    responsiveContext =>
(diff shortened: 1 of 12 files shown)


src/components/Responsive/Visible/index.js
```diff
@@ -0,0 +1,48 @@
+import React, { useCallback } from 'react';
+import PropTypes from 'prop-types';
+import intersection from 'lodash/intersection';
+
+import Context from '../Context';
+
+const Visible = ({ children, breakpoints, ...passedProps }) => {
+  const getIsVisible = useCallback(
+    responsiveContext => !!intersection(breakpoints, responsiveContext).length,
```
rheckart: I tried using the `!!` over in tk world and received a lint exception for it (Redundant double negation.eslint(no-extra-boolean-cast). Is it ok to use this over here?

rheckart: :man_shrugging: Ya I didn't understand why it would warn me (like it was a bad idea to use it). Anyhow, approved the request.

hpieroni: I think it's clearer if you just use `lodash/isEmpty`:

```suggestion
    responsiveContext => !isEmpty(intersection(breakpoints, responsiveContext)),
```

PR #105, opened 2019-09-16
WIP // Avatar Component

chelshaw:
First pass at adding Avatar to our components. Creating this pre-emptive PR for feedback and throughts

I'm treating Avatar as a block that can have child addons inside of it (that's the spots for badges, etc). The badges themselves are controlled by the parent, but there is potential here for passing contexts (eg. sizing) and making separate AvatarBadge components to keep the styling consistent.

The size takes either an accepted list of strings, or number of pixels. You'll notice a lot of generating styles based on the props... I wasn't sure the best elegant way to do this without using styles. Open to all feedback! And i'm aware I still need a README :)


src/components/Avatar/Addon.js
```diff
@@ -0,0 +1,35 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import style from './style.css';
+
+const Addon = ({ position, offset, ...rest }) => {
+  const positions = useMemo(() => {
+    const p = position.split('-');
+    return {
+      [p[0]]: `${offset}px`,
```
pixelbandito: What if you just let the user override position with the `rest` prop in `Addon`?

Whitelisting just the 4 corner names could backfire. A future dev could get `top-right` and `right-top` confused.

It also feels a bit arbitrary to use the same offset for x and y. There could be cases where a consumer needs to alter the position a bit to accommodate for non-circular add-on content.

You could still provide a default of `{ bottom: '0', right: '0' }`

src/components/Avatar/Addon.js
```diff
@@ -0,0 +1,35 @@
...
+      [p[1]]: `${offset}px`,
+    };
+  }, [position, offset]);
+
+  const badgeClassNames = classNames(style.AvatarBadge, rest.className);
+  return <div {...rest} className={badgeClassNames} style={positions} />;
```
pixelbandito: If you keep the `positions` props, would you change this to ...
```
{{
  ...positions,
  rest.style,
}}
```
... so the consumer doesn't lose the ability to customize or override styles.

src/components/Avatar/index.js
```diff
@@ -0,0 +1,82 @@
+import React, { useMemo } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import Addon from './Addon';
+import style from './style.css';
+
+const semanticSizes = {
+  small: 40,
```
pixelbandito: Can we import the base sizes and use multiples of them here?

src/components/Avatar/style.css
```diff
@@ -0,0 +1,42 @@
...
+  position: relative;
+}
+
+.AvatarBadge {
+  position: absolute;
+  bottom: 0;
+  right: 0;
+}
+
+@keyframes imagePlaceholderShimmer {
```
pixelbandito: How about putting the keyframes in a global stylesheet? "Placeholder shimmer" could even be its own component.

src/components/Avatar/Addon.js
chelshaw: The expectation is that it's always used in Avatar, which is always a circle. Happy to allow style to overwrite those settings. Also, i could whitelist the reversed names as well, the code will work for either. Curious what @mdespuits thinks about the approaches

mdespuits: I feel like allowing overriding of arbitrary styles isn't really necessary for most cases. I like the idea of whitelisting the reverse options in case there is some confusion, but I tend to fall on the side of keeping small components like this closed and opening up changes as needed.

So unless there is a need for overriding styles instead of using CSS classes I'm fine with your approach so long as the others are whitelisted. However, there could be the argument that if they type it backwards they will get console proptype errors and it won't work as they expected, so they will probably figure out their error before it gets anywhere close to a live system.

PR #120, opened 2019-10-17
Add and fix tests for Avatar and Button

chelshaw:
Added more substantial tests for Avatar. Was throwing some proptype errors for size on Addon, which is a required proptype but not supplied in the tests. Not sure if it should actually be required or have a default value, open to feedback there (and everywhere)


pixelbandito:
Adding a default size seems like a win-win, right?

PR #124, opened 2019-10-21
Add social icons to Icon component

nvmusoke:
updated existing icons & created new icons using icons from designer provided folder.


mdespuits:
Before I look closely at the existing Icon components, it's worth mentioning that this may be a good opportunity to talk about using `svgo` as a tool to optimize our SVG icons: https://github.com/svg/svgo. Just browsing through some of these I'm fairly sure incorporating some optimization will be a good practice and also possible to add to testing/committing workflow.

@chelshaw @pixelbandito thoughts?

src/components/Icon/icons/Twitter.js
```diff
@@ -7,11 +7,14 @@ const Twitter = ({ fill, ...props }) => (
...
...42188,18.4628906 C2.64778604,18.4628906 2.56640664,18.4596354 2.48828125,18.453125 C2.40364541,18.4466146 2.32063843,18.4401042 2.23925781,18.4335938 C2.1578772,18.4270833 2.07812539,18.4173178 2,18.4042969 C2.45573145,18.6972671 2.93098711,18.9576812 3.42578125,19.1855469 C3.92057539,19.4134126 4.43163799,19.605468 4.95898438,19.7617188 C5.48633076,19.92448 6.02831753,20.0465491 6.58496094,20.1279297 C7.14160435,20.2093103 7.70963252,20.25 8.2890625,20.25 C10.1770928,20.25 11.8469979,19.8968134 13.2988281,19.1904297 C14.7506583,18.4840459 15.9713492,17.5709691 16.9609375,16.4511719 C17.9440153,15.3313746 18.6910782,14.0781319 19.2021484,12.6914062 C19.7132187,11.3046806 19.96875,9.93425156 19.96875,8.58007812 C19.96875,8.53450498 19.9671224,8.48893252 19.9638672,8.44335938 C19.960612,8.39778623 19.9589844,8.35221377 19.9589844,8.30664062 L19.9589844,8.1796875 C19.9589844,8.13411436 19.9557292,8.08854189 19.9492188,8.04296875 C20.1510427,7.89973887 20.3447256,7.74674561 20.5302734,7.58398438 C20.7158212,7.42122314 20.8964835,7.251954 21.0722656,7.07617188 C21.2415373,6.90038975 21.4042961,6.71484473 21.5605469,6.51953125 C21.7167977,6.32421777 21.8632806,6.12565205 22,5.92382812 Z"
```
mdespuits: Perfect example of how svgo could help. This is a very long `d` for the Twitter icon. Running this through svgo, you can trim this same icon to 7% of this size: 
image

nvmusoke: ooh cool! I'm looking into this now.

pixelbandito:
@nvmusoke I'd be happy to approve this, but I agree that you should run those icons through svgo first. I remember it being pretty easy, but let me or @mdespuits know if you need a hand.

nvmusoke:
@mdespuits @pixelbandito yeah i'm gonna run them through svgo. I tested it in a new branch and it was really simple.

src/components/Icon/icons/Google.js
```diff
@@ -0,0 +1,44 @@
...
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="24"
+    height="24"
+    viewBox="0 0 24 24"
+    {...props}
+  >
+    <g fill="none" fillRule="evenodd" transform="translate(3 3)">
+      <path
+        fill={fill}
```
chelshaw: I think these need to be `{fill[0]}` `{fill[1]}` etc since it's an array

src/components/Icon/icons/Google.js
```diff
@@ -0,0 +1,44 @@
...
+      />
+      <path
+        fill={fill}
+        d="M12 6.58c1.321 0 2.508.454 3.44 1.345l2.582-2.58C16.463 3.891 14.426 3 12 3a8.997 8.997 0 00-8.043 4.958l3.007 2.332C7.672 8.163 9.656 6.58 12 6.58z"
+      />
+    </g>
+  </svg>
+);
+
+Google.propTypes = { fill: PropTypes.string };
```
chelshaw: and this should expect an array of strings, not just a single string

src/components/Icon/icons/Reddit.js
```diff
@@ -11,10 +11,16 @@ const Reddit = ({ fill, ...props }) => (
...
-    />
+    <g fill="none" fillRule="evenodd">
+      <path
+        fill="#FF4500"
```
chelshaw: We should remove this and set it as default... and maybe the white bit is always white? open to thoughts

src/components/Icon/icons/Slack.js
```diff
@@ -0,0 +1,32 @@
...
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="24"
+    height="24"
+    viewBox="0 0 24 24"
+    {...props}
+  >
+    <g fill="none">
+      <path
+        fill="#E01E5A"
```
chelshaw: We can use the mapping technique here too, like you did with google's icon

src/components/Icon/icons/Google.js
nvmusoke: done

nvmusoke:
> A few updates
> 
> What are all the files in /Icon/main for?

All the svgs i've yet to make js files for. I initially added them because i thought I needed them, but then I thought it would be easier to keep them in the same project.

pixelbandito:
My initial preference is not to keep working .svg files in the codebase, but I can see how it's a helpful way to keep track of what's left to do.

I'm ok with it.

PR #125, opened 2019-10-22
Correctly setup re-renders for performance

mdespuits:
When trying to setup an `<input />` with value state wrapped inside of a themed component, the entire tree inside of the themed components would re-render. For example:

```jsx
const Test = () => {
  const [val, setVal] = useState('');
  return (
    <Media>
      <input type="text" value={val} onChange={e => setVal(e.target.value)} />
    </Media>
  );
};
```

When a user changed the value of the text input, the entire `Media` tree would re-render. Thanks to one small change to `withDefaultTheme`, this no longer happens and the added test for `Media` ensures that this regression will not occur again.

### DIFF
--- a/src/components/Media/test.js
+++ b/src/components/Media/test.js
@@ -1,7 +1,28 @@
-import React from 'react';
+import React, { useState } from 'react';
 import { mount } from 'enzyme';
 
 import Media from './';
+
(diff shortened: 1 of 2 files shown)


pixelbandito:
I like how you made the test check for the actual bug we found!

PR #130, opened 2019-11-12
Update button color, add text-decoration none, rename CSS variable

pixelbandito:
This is a **patch version**, since it doesn't add any functionality, it just fixes a bug and cleans up internal workings.

Updated primary button color to Cision green-ish blue. (Bug fix)

`text-decoration: none` for when a button uses an `a` or `Link` as the tag. (Bug fix)

CSS var name cleanup (`--rvr-rvr-*` -> `--rvr-*`)

Made example app use the same package versions as the parent by default, hopefully that will squash some of the security warnings on the github repo.

Added yalc to .gitignore. I used yalc to test locally, but I'm not confident enough yet to add instructions to the README. I will once I have the process down.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,5 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 storybook-static
+.yalc
+yalc.lock
--- a/package.json
+++ b/package.json
@@ -59,10 +59,10 @@
     "deploy": "gh-pages -d docs",
     "storybook": "start-storybook -p 9009",
     "build-storybook": "build-storybook -o docs",
-    "deploy:storybook": "yarn build-storybook && yarn gh-pages -d docs",
-    "prop-types": "^15.5.4"
+    "deploy:storybook": "yarn build-storybook && yarn gh-pages -d docs"
   },
   "peerDependencies": {
+    "prop-types": "^15.5.4",
     "react": "^16.9.0",
     "react-dom": "^16.9.0"
   },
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -59,12 +59,12 @@
     "deploy": "gh-pages -d docs",
     "storybook": "start-storybook -p 9009",
     "build-storybook": "build-storybook -o docs",
-    "deploy:storybook": "yarn build-storybook && yarn gh-pages -d docs"
+    "deploy:storybook": "yarn build-storybook && yarn gh-pages -d docs",
+    "prop-types": "^15.5.4"
```
mdespuits: This is in the wrong place. `yarn prop-types`?

.gitignore
```diff
@@ -25,3 +25,5 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 storybook-static
+.yalc
+yalc.lock
```
mdespuits: What is `yalc`?

package.json
pixelbandito: oh, right that is a real dependency. Whoops!

.gitignore
pixelbandito: `yalc` solves the same problem as `yarn link/npm link`, but it actually copies deps and maintains a sort of local package registry.

https://github.com/whitecolor/yalc

It works more like real dependencies, and, in this case, solved some issues with peer deps when I was trying to test local front-ends importing local RoverUI.

However, I haven't figured out the best options to use it with, so I had to let it make changes to my front-end code's package.json file. Not ideal, long term, so I haven't documented it as a "solution" for local testing.

I wanted to add it to .gitignore, because I anticipate using it for testing again and this reduces the risk of me pushing local registry metadata to git.

package.json
pixelbandito: Done!

PR #132, opened 2019-11-18
B and C named icons

nvmusoke:
still going through adding icons. Finished everything starting with a B and C.


src/components/Icon/icons/ChartBars2.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import iconColorsMap from './colors';
+
+const ChartBars2 = ({ fill, ...props }) => (
```
chelshaw: Wondering if we should just call it `ChartBars` if we're not also importing the presumed original `ChartBars` ?

chelshaw: Or maybe name it descriptively like the others

nvmusoke: I was just going off the names given to me in the svg file, but i agree that a few of them could have better names and happy to change any.

pixelbandito: I second @chelshaw ! Otherwise looks good.

pixelbandito:
FYI: This should be a minor version bump. I think it's fine to do the version bump in this PR, then wait until it's merged before `yarn publish`ing from master.

PR #133, opened 2019-11-18
Added Button level 'text' and added yalc note to README.md

pixelbandito:
Added .yalc instructions with a big warning to the README.

Added a "Text" level to button that looks like blue text, no border. I did this because:
1. I noticed Invision had a button like that in the stylesheet.
2. Bootstrap and Foundation both have buttons like that, but theirs are called "Link"s, both preserve the padding of normal buttons, and one has an underline on hover.
3. In Zeplin, Egle had two borderless buttons, one called "Text/Blue" and one called "Text/Gray", and without padding.

I figured "text" was a little less confusing than "link" as a modifier, since these don't get typical underline styles.
I also figured I'd leave the padding / sizing as it is, following Bootstrap/Foundation, and ppl can override the styles with `padding: 0` when/if necessary.
I don't know what to do as far as naming convention to add the gray and blue variations. (Request for comments)

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -10,6 +10,7 @@ import style from './style.css';
 
 const Button = props => {
   const {
+    active,
     children: initChildren,
     className: passedClassName,
     darkMode,
@@ -28,6 +29,7 @@ const Button = props => {
     style[size],
     {
       [style.darkMode]: darkMode,
+      [style.active]: active,
     }
   );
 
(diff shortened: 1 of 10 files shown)


src/components/Button/style.css
```diff
@@ -145,6 +145,29 @@
   border-color: var(--rvr-border-color-tertiary-disabled);
 }
 
+.text,
+.text:link,
+.text:visited {
+  color: var(--rvr-color-text);
```
mdespuits: Instead of re-declaring the properties and the values, you should be able to declare a component variable that you can re-assign in the scope. For example:

```css
.text {
  --rvr-btn-color: var(--rvr-color-text);
  color: var(--rvr-btn-color); /* sets the default */
}

.text:active {
  /* Changes the variable for the scope, but no property declaration necessary */
  --rvr-btn-color: var(--rvr-color-text-active);
}
```

mdespuits: Also, this is not a blocker, but more of a bikeshed thought about how to build out these variants

src/components/Button/index.js
```diff
@@ -56,19 +56,24 @@ const Button = props => {
   );
 };
 
+export const levels = [
+  'primary',
+  'secondary',
+  'success',
+  'teal',
+  'teal',
```
mdespuits: Double `teal`

src/components/Button/style.css
pixelbandito: I like this idea, but without spending some time trying to dedupe styles, it doesn't seem like an obvious improvement in this component (since the variables-to-style rules ratio is 1:1).

I'll think about approaches to drying this up, and If I'm missing something, please let me know. For now, I'd like to merge as-is.

PR #134, opened 2019-11-18
made selected menu tab darker and bolder. re arranged icon in iconlist

nvmusoke:
**SUMMARY:** Per https://jira.cision.com/browse/EVER-3594, made active tab name color darker, and attempted to make it bolder but looks like that particular font doesn't get bold. Also went ahead and re ordered an Icon in the icon list so it's in alphabetical order. 
**RISK:** low
**VERIFICATION:** in roverUI, the `EasyTabMenu` overview should have a noticeably darker active tab.


mdespuits:
@nvmusoke Since this is a visible change, can you provide a screenshot showing the changes?

src/components/TabMenu/Item/style.css
```diff
@@ -22,7 +22,7 @@
 }
 
 .active {
-  color: var(--rvr-gray);
-  font-weight: 700;
+  color: var(--rvr-gray-90);
+  font-weight: 800;
```
pixelbandito: I just checked in the storybook app - looks like the .active class font-weight is being overridden by the Button font weight.

The `.active` style gets added to an `li` in `TabMenu/Item`, but the `button` element is inside the `li`, and it has it's own font-weight.

I can help you figure out how to get an `.active` class onto the button if that would help.

nvmusoke:
New version:
Screen Shot 2019-11-19 at 11 56 57 AM

PR #135, opened 2019-12-02
Callout

### LINKED ISSUE #131: Add StatusBar to rover-ui as Callout
image

mdespuits:
A fairly small component that will draw attention to users for warnings or information that they either need to know or need to act on.

Fixes #131

### DIFF
--- a/src/components/Callout/style.css
+++ b/src/components/Callout/style.css
@@ -1,33 +1,29 @@
 .Callout {
-  --rvr-callout-bg: var(--rvr-blue-lite-2);
-  --rvr-callout-border-color: var(--rvr-blue);
-  --rvr-callou-color: var(--rvr-gray-80);
-
   border-radius: 2px;
   box-sizing: border-box;
   display: flex;
   padding: 20px;
   position: relative;
-  background-color: var(--rvr-callout-bg);
-  border-bottom-color: var(--rvr-callout-border-color);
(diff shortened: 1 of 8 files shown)


src/components/Callout/style.css
```diff
@@ -0,0 +1,63 @@
...
+  position: relative;
+  background-color: var(--rvr-callout-bg);
+  border-bottom-color: var(--rvr-callout-border-color);
+  border-bottom-style: solid;
+  border-bottom-width: 1px;
+  color: var(--rvr-callout-color);
+}
+
+.success {
+  --rvr-callout-bg: var(--rvr-green-lite-2);
```
pixelbandito: We might want to sanity check that this works in IE11. I know custom css props have given us troubles occasionally.

src/components/Icon/story.js
```diff
@@ -14,22 +17,47 @@ storiesOf('Planets/Icon', module)
     },
   })
   .add('Overview', () => (
-    <Icon name={select('name', iconNames, 'times-circle')} />
+    <Icon name={select('name', iconNames, iconNames[0])} />
```
pixelbandito: Nice enhancements here.

src/components/Callout/style.css
mdespuits: That is actually a very good point. I'll take a look later this morning

PR #137, opened 2019-12-20
converted all svgs, removed all original svg files

nvmusoke:
converted the remaining svgs using hygen injectors. did a visual check and they seem good. I'll still need to go back and work on some tricky alignment in the examples.


nvmusoke:
Screen Shot 2019-12-20 at 4 00 38 PM

PR #139, opened 2019-12-31
Migrate to a fork of react-container-query ...

pixelbandito:
... that doesn't have deprecated lifecycle methods, but has new interface.

Now you have to do
```
<ContainerQuery query={query}>
      {(params, ref) => (
        <div ref={ref}>
          {/* Whatever */}
        </div>
      }
</ContainerQuery>
```

In other words, you should explicitly put a ref on the element you're measuring.

### DIFF
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -9,52 +9,52 @@
   dependencies:
     "@babel/highlight" "^7.0.0"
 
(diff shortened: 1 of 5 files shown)


pixelbandito:
I need to bump the version number on this, otherwise I believe it's good to go.

PR #144, opened 2020-02-03
Fixes #141 - Remove grid style from EasyDropdown toggle, ...

### LINKED ISSUE #141: Button display is set to Grid
Screen Shot 2020-01-06 at 2 34 15 PM

pixelbandito:
... improve logic around wrapping children in a Button, add flex style for Button body string when Button.Addon is present

### DIFF
--- a/src/components/EasyDropdown/index.js
+++ b/src/components/EasyDropdown/index.js
@@ -53,21 +53,28 @@ const EasyDropdown = ({
   );
 
   const children = useMemo(() => {
-    return typeof initChildren === 'string' ? (
-      <Button
-        className={classNames(toggleProps.className, style.toggle)}
-        data-is-open={isOpen}
-        onClick={onToggle}
-      >
+    const childrenArray = React.Children.toArray(initChildren);
+    const hasStringChildren = childrenArray.reduce(
+      (result, child) => result || typeof child === 'string',
+      false
+    );
+
+    /*
+      If any children are raw strings, wrap all the children in a <Button />
+      Otherwise, assume there's a custom toggle component, and clone it with
+      added props for behavior.
+    */
+    return hasStringChildren ? (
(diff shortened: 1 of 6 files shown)


src/components/EasyDropdown/index.js
```diff
@@ -53,21 +53,28 @@ const EasyDropdown = ({
...
+      false
+    );
+
+    /*
+      If any children are raw strings, wrap all the children in a <Button />
+      Otherwise, assume there's a custom toggle component, and clone it with
+      added props for behavior.
+    */
+    return hasStringChildren ? (
+      <Button {...toggleProps} data-is-open={isOpen} onClick={onToggle}>
```
nvmusoke: just for my own education, what is data-is-open?

pixelbandito: That's an html [data attribute](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes), which has no effect by itself. 
I'm using it in the unit test for the component. 
Think of them as a catch-all place to put custom data. They're frequently used to give automated tests hooks to find things.

src/components/EasyDropdown/story.js
```diff
@@ -96,6 +96,26 @@ storiesOf('Galaxies/EasyDropdown', module)
...
+            Button with addon
+            <Button.Addon>→</Button.Addon>
+          </EasyDropdown>
+          <br />
+          <br />
+          <EasyDropdown
+            menuItems={[{ label: 'Click me!', onClick: () => {} }]}
+            defaultIsOpen={false}
+            style={{ width: '100%' }}
+            toggleProps={{ style: { width: '100%' } }}
```
pixelbandito: I don't love the need to declare `width: 100%` twice, but I didn't have any better ideas, considering the `EasyDropdown` has both it's own wrapping div _and_ an element for the toggle button.

PR #146, opened 2020-02-10
fix: validate image asynchronously by HTMLImageElement instead of XHR

### LINKED ISSUE #145: Avatar component not rendering images
The component decides whether to render an image or not by performing a`HEAD` request to the given `imageUrl`, if the server responds with a `200` the image will be rendered. The problem is that if the server doesn't allow `HEAD` requests, images won't be rendered even though they are valid image URLs.

Another problem is that this **URL check process** is blocking the main thread.

sebastianvera:
Use an HTMLImageElement instance to validate image URLs instead of
checking the response of a HEAD request. The problem with that approach
is that if the servers don't allow HEAD requests, no images will be
loaded even though they are valid URLs. This solution also performs this
validation process asynchronously so no more blocking the main thread by
performing XHR requests.

Fixes #145

### DIFF
--- a/src/components/Avatar/index.js
+++ b/src/components/Avatar/index.js
@@ -15,6 +15,8 @@ const semanticSizes = {
 // This function will be used in a useEffect so make sure it returns a cleanup function
 function defaultImageFetcher({ onLoad, onError, src }) {
   const img = new Image();
+  img.referrerPolicy = 'no-referrer';
+  img.crossOrigin = 'Anonymous';
   img.onload = function() {
     if ('naturalHeight' in this) {
       if (this.naturalHeight + this.naturalWidth === 0) {
--- a/src/components/Avatar/index.js
+++ b/src/components/Avatar/index.js
@@ -37,7 +37,7 @@ const Avatar = ({
     };
   }, [size]);
 
-  const [loadedImageUrl, setLoadedImageUrl] = useState(null);
+  const [loaded, setLoaded] = useState(false);
 
   const initials = useMemo(() => {
     if (!name) return '';
@@ -52,7 +52,7 @@ const Avatar = ({
   useEffect(() => {
     if (!imageUrl) return () => {};
 
-    setLoadedImageUrl(null);
+    setLoaded(false);
 
     const img = new Image();
     img.onload = function() {
@@ -64,10 +64,10 @@ const Avatar = ({
         return;
       }
 
-      setLoadedImageUrl(imageUrl);
+      setLoaded(imageUrl);
     };
     img.onerror = function() {
-      setLoadedImageUrl(null);
+      setLoaded(true);
     };
     img.src = imageUrl;
 
@@ -81,8 +81,7 @@ const Avatar = ({
 
   const mainStyle = {
     ...sizes,
-    backgroundImage:
-      loadedImageUrl && !loading ? `url(${loadedImageUrl})` : undefined,
+    backgroundImage: loaded && !loading ? `url(${imageUrl})` : undefined,
     ...rest.style,
   };
 
@@ -94,7 +93,7 @@ const Avatar = ({
       {...rest}
       style={mainStyle}
     >
-      {!loading && !loadedImageUrl && initials}
+      {!loading && !loaded && initials}
       {children}
     </div>
   );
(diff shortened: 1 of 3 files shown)


src/components/Avatar/index.js
```diff
@@ -37,6 +37,8 @@ const Avatar = ({
     };
   }, [size]);
 
+  const [loadedImageUrl, setLoadedImageUrl] = useState(null);
```
pixelbandito: We don't care about the actual url, do we?

What do you think of renaming this to `loaded` and making it a boolean?

src/components/Avatar/index.js
```diff
@@ -47,22 +49,40 @@ const Avatar = ({
...
-      return http.status === 200 ? imageUrl : null;
-    } catch (e) {
-      return null;
-    }
+  useEffect(() => {
+    if (!imageUrl) return () => {};
+
+    setLoadedImageUrl(null);
+
+    const img = new Image();
```
pixelbandito: I would have probably put the Image in a `ref` and used `removeEventListener` and a named function for the event handler.

But I can't think of a reason this wouldn't work, unless the anonymous function's closure does something weird with the state setter `setLoadedImageUrl`. I'll approve it, as long as you'll keep an eye out for weird side effects.

Maybe add a unit test for the case where the `imageUrl` changes after mounting the component?

src/components/Avatar/index.js
```diff
@@ -47,22 +49,40 @@ const Avatar = ({
...
-    try {
-      const http = new XMLHttpRequest();
-      http.open('HEAD', imageUrl, false);
-      http.send();
-      return http.status === 200 ? imageUrl : null;
-    } catch (e) {
-      return null;
-    }
+  useEffect(() => {
+    if (!imageUrl) return () => {};
```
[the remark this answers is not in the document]
sebastianvera: empty return works, but eslint was complaining about that (`consistent-return` rule)

src/components/Avatar/index.js
sebastianvera: Didn't quite get the ref solution but I'm taking care of weird state updates (caused by race conditions) by removing the `onload` and `onerror` callbacks in the cleanup function every time `imageUrl` changes. I can add a new test case, but I'll need to add two new props `onLoad` and `onError`, so it can be easily tested 🙂

src/components/Avatar/index.js
sebastianvera: agreed! updated

src/components/Avatar/index.js
sebastianvera: I'm not sure how to test it, I'll need to add a delay so I can change the `imageUrl` prop before `onload` gets called, any idea on how to test that?

pixelbandito: I think you're ok as it is. I can see that being a pain to test.
You were at least able to validate in storybook, right?

sebastianvera: I did validate it in Storybook. Anyway, I figured out a way of testing it, let me know if you like it, otherwise, we revert it.

sebastianvera: cc: @pixelbandito

pixelbandito:
lgtm. Thanks for the extra test!

sebastianvera:
@pixelbandito How should we proceed? I cannot merge this

pixelbandito:
Hey @sebastianvera, there are a couple of small extra things you could do.

1. Run a local test importing your version of RoverUI's Avatar into the Cision React app's front-end. This isn't strictly necessary since your changes will be in a new version, but it's helpful.
2. Bump the version and tag in your branch.

Then I'll take care of merging to master and publishing to npm.

I just opened a branch with some updated instructions on the README that should be helpful.
[Yalc testing](https://github.com/cision/rover-ui/blob/readme-update-tag-version-tweak-yalc-notes/README.md#user-content-testing-in-monorepos-with-yalc-experimental)
[Version bumps](https://github.com/cision/rover-ui/blob/readme-update-tag-version-tweak-yalc-notes/README.md#user-content-publishing-a-new-version)

sebastianvera:
@pixelbandito Tested it locally and bumped the version, looking good 🙂

PR #148, opened 2020-02-16
Correct pill font

kcj002:
Pill font styling was Source Sans Pro, which was TrendKite styling. Post rebrand, Source Sans Pro was replaced with Proxima Nova. Removed styling for Menu.Item and Pill so that global font of Proxima Nova would be used.

Risk: low, just a font update.

### DIFF
--- a/src/shared/variables.css
+++ b/src/shared/variables.css
@@ -4,6 +4,7 @@
 
   /* Fonts */
   --rvr-base-font-color: #58666e;
+  --rvr-base-font-family: 'Proxima Nova', sans-serif;
   --rvr-font-family-mono: SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
 
   /* Transitions */
--- a/src/shared/variables.css
+++ b/src/shared/variables.css
@@ -4,7 +4,6 @@
 
   /* Fonts */
   --rvr-base-font-color: #58666e;
-  --rvr-base-font-family: 'Source Sans Pro', sans-serif;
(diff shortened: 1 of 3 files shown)


pixelbandito:
Hey, @kcj002 any reason not to put the new font in the `variables.css` instead?
Without a font-family variable, we'll be relying on the hosting app to handle that part of the style.

kcj002:
I thought the `font-family: 'Proxima Nova', sans-serif;` for :root in colors.css would cover font assignment. Is that incorrect?

pixelbandito:
> I thought the `font-family: 'Proxima Nova', sans-serif;` for :root in colors.css would cover font assignment. Is that incorrect?

Ah, I missed that because it's not supposed to be in `colors.css`. Would you mind moving it to `variables.css` as part of your change?

src/shared/variables.css
```diff
@@ -1,10 +1,11 @@
 :root {
+  font-family: 'Proxima Nova', sans-serif;
```
pixelbandito: Sorry I didn't realize that was a style rule in `colors.css`  - I thought it was a CSS variable.
Safe to just omit it, but it shouldn't hurt anything.

kcj002: So I can remove this from variables and colors.css and then recreate the font variable and use that for Menu.Item and Pill

pixelbandito: Yes, exactly. If/when we want to apply a `font-family` to a component, we should re-add it to `variables.css`

If you want to do that now, great, but it can also wait if necessary.

PR #155, opened 2020-04-02
Component: Tooltip

mdespuits:
Finally, a way to add a basic tooltip for text as well as the configurability to display any content we want.

image
image

### DIFF
--- a/src/components/Tooltip/index.js
+++ b/src/components/Tooltip/index.js
@@ -74,7 +74,6 @@ const Tooltip = ({
     style: {
       width: (tooltipStyle && tooltipStyle.width) || width,
       ...offsets,
-      ...(tooltipOpts.style || {}),
     },
     className: tooltipClassNames,
   };
--- a/src/components/Tooltip/index.js
+++ b/src/components/Tooltip/index.js
@@ -1,0 +1,139 @@
+import React, { useRef, useEffect, useState, useMemo } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import styles from './style.css';
+
+import Icon from '../Icon';
+
+export const directions = ['top', 'left', 'right', 'bottom'];
+
+const Tooltip = ({
+  children,
+  closeable,
+  content: tooltipContent,
+  direction,
+  hover,
+  tooltipOpts: tooltipOptsProp,
+  open,
+  onClose,
+  width,
+  ...rest
+}) => {
+  const childrenRef = useRef(null);
+  const tooltipRef = useRef(null);
+
+  const [hovered, setHovered] = useState(false);
+  const [childrenWidth, setWidth] = useState({});
+  const [childrenHeight, setHeight] = useState(0);
+
+  useEffect(() => {
+    if (childrenRef.current) {
+      setHeight(childrenRef.current.offsetHeight);
+      setWidth(childrenRef.current.offsetWidth);
+    }
+  });
+
+  const handleSetHover = value => () => {
+    if (hover) {
+      setHovered(value);
+    }
+  };
+
+  const offsets = useMemo(() => {
+    if (direction === 'top') {
+      return {
+        bottom: childrenHeight + 20,
+      };
+    } else if (direction === 'bottom') {
+      return {
+        top: childrenHeight + 20,
+      };
+    } else if (direction === 'right') {
+      return {
+        left: childrenWidth + 20,
+      };
+    }
+    return {
+      right: childrenWidth + 20,
+    };
+  });
+
+  const {
+    style,
+    className: tooltipClassName,
+    ...tooltipOpts
+  } = tooltipOptsProp;
+
+  const tooltipProps = {
+    ...tooltipOpts,
+    style: {
+      ...offsets,
+      width: (style && style.width) || width,
+    },
+    className: classNames(styles.Tooltip, styles[direction], tooltipClassName, {
+      [styles.open]: open || hovered,
+    }),
+  };
+
+  const TooltipContent = tooltipContent && (
+    <div {...tooltipProps}>
+      <div style={style} className={styles.tooltipInnerWrapper}>
+        {closeable && (
+          <button className={styles.tooltipClose} onClick={onClose}>
+            <Icon name="close" />
+          </button>
+        )}
+        <div ref={tooltipRef}>{tooltipContent}</div>
+      </div>
+    </div>
+  );
+
+  return (
+    <div className={styles.wrapper} {...rest}>
+      {TooltipContent}
+      <div
+        ref={childrenRef}
+        onMouseEnter={handleSetHover(true)}
+        onMouseLeave={handleSetHover(false)}
+        className={styles.original}
+      >
+        {children}
+      </div>
+    </div>
+  );
+};
+
+Tooltip.propTypes = {
+  children: PropTypes.node.isRequired,
+  closeable: PropTypes.bool,
+  content: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.string,
+    PropTypes.element,
+  ]),
+  direction: PropTypes.oneOf(directions),
+  hover: PropTypes.bool,
+  onClose: PropTypes.func,
+  open: PropTypes.bool,
+  text: PropTypes.string,
+  tooltipOpts: PropTypes.shape({
+    className: PropTypes.string,
+    style: PropTypes.object,
+  }),
+  width: PropTypes.string,
+};
+
+Tooltip.defaultProps = {
+  content: null,
+  closeable: false,
+  direction: 'top',
+  hover: false,
+  onClose: () => {},
+  open: false,
+  text: '',
+  tooltipOpts: {},
+  width: '150px',
+};
+
+export default Tooltip;
--- a/src/components/Tooltip/index.js
+++ b/src/components/Tooltip/index.js
@@ -43,7 +43,7 @@ const Tooltip = ({
   });
 
   const {
-    style,
+    style: tooltipStyle,
     className: tooltipClassName,
     ...tooltipOpts
   } = tooltipOptsProp;
@@ -72,7 +72,7 @@ const Tooltip = ({
   const tooltipProps = {
     ...tooltipOpts,
     style: {
-      width: (style && style.width) || width,
+      width: (tooltipStyle && tooltipStyle.width) || width,
       ...offsets,
       ...(tooltipOpts.style || {}),
     },
@@ -81,7 +81,7 @@ const Tooltip = ({
 
   const TooltipContent = (text || tooltipContent) && (
     <div ref={tooltipRef} {...tooltipProps}>
-      <div style={style} className={styles.tooltipInnerWrapper}>
+      <div style={tooltipStyle} className={styles.tooltipInnerWrapper}>
         {closeable && (
           <button className={styles.tooltipClose} onClick={onClose}>
             <Icon name={closeIcon} />
--- a/src/components/Tooltip/index.js
+++ b/src/components/Tooltip/index.js
@@ -1,4 +1,10 @@
-import React, { useRef, useEffect, useState, useMemo } from 'react';
+import React, {
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
@@ -17,7 +23,6 @@ const Tooltip = ({
   hover,
   tooltipOpts: tooltipOptsProp,
   open,
-  onClose,
   text,
   width,
   ...rest
@@ -28,6 +33,8 @@ const Tooltip = ({
   const [hovered, setHovered] = useState(false);
   const [tooltipHeight, setTooltipHeight] = useState(0);
   const [childrenHeight, setChildrenHeight] = useState(0);
+
+  const closeFunc = useCallback(closeable || function() {}, []);
 
   const handleSetHover = value => () => {
     if (hover) {
@@ -40,7 +47,7 @@ const Tooltip = ({
       setTooltipHeight(tooltipRef.current.offsetHeight);
       setChildrenHeight(childrenRef.current.offsetHeight);
     }
-  });
+  }, []);
 
   const {
     style: tooltipStyle,
@@ -81,8 +88,8 @@ const Tooltip = ({
   const TooltipContent = (text || tooltipContent) && (
     <div ref={tooltipRef} {...tooltipProps}>
       <div style={tooltipStyle} className={styles.tooltipInnerWrapper}>
-        {closeable && (
-          <button className={styles.tooltipClose} onClick={onClose}>
+        {!!closeable && (
+          <button className={styles.tooltipClose} onClick={closeFunc}>
             <Icon name={closeIcon} />
           </button>
         )}
@@ -114,7 +121,7 @@ const Tooltip = ({
 
 Tooltip.propTypes = {
   children: PropTypes.node.isRequired,
-  closeable: PropTypes.bool,
+  closeable: PropTypes.func,
   closeIcon: PropTypes.string,
   content: PropTypes.oneOfType([
     PropTypes.node,
@@ -123,7 +130,6 @@ Tooltip.propTypes = {
   ]),
   direction: PropTypes.oneOf(directions),
   hover: PropTypes.bool,
-  onClose: PropTypes.func,
   open: PropTypes.bool,
   text: PropTypes.string,
   tooltipOpts: PropTypes.shape({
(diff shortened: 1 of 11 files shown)


src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+      </div>
+    </div>
+  );
+};
+
+Tooltip.propTypes = {
+  children: PropTypes.node.isRequired,
+  closeable: PropTypes.bool,
+  closeIcon: PropTypes.string,
+  content: PropTypes.oneOfType([
```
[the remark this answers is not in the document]
mdespuits: Nice. I did not know this. Thanks!

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+    PropTypes.node,
+    PropTypes.string,
+    PropTypes.element,
+  ]),
+  direction: PropTypes.oneOf(directions),
+  hover: PropTypes.bool,
+  onClose: PropTypes.func,
+  open: PropTypes.bool,
+  text: PropTypes.string,
```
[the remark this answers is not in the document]
mdespuits: Yeah, I did do that. I switched over to this one, but I wasn't convinced either way. I didn't like doing the switch in either case. I'm happy to go back to a `typeof` check

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+      </div>
+    </div>
+  );
+
+  return (
+    <div className={styles.wrapper} {...rest}>
+      {TooltipContent}
+      <div
+        ref={childrenRef}
+        onMouseEnter={handleSetHover(true)}
```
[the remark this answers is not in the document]
mdespuits: Hmm, I think I can manage that

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+    style: {
+      width: (tooltipStyle && tooltipStyle.width) || width,
+      ...offsets,
+    },
+    className: tooltipClassNames,
+  };
+
+  const TooltipContent = (text || tooltipContent) && (
+    <div ref={tooltipRef} {...tooltipProps}>
+      <div style={tooltipStyle} className={styles.tooltipInnerWrapper}>
```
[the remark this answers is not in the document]
mdespuits: Actually, no. To accomplish the tooltip arrow's "shadow", the z-index needed to be defined on an element nested within the outer ref. I'd have to walk through it again in a video to remind myself of why that was, but it wasn't possible without an extra nesting layer ☹️

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+        {closeable && (
+          <button className={styles.tooltipClose} onClick={onClose}>
+            <Icon name={closeIcon} />
+          </button>
+        )}
+        <div
```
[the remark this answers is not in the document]
mdespuits: Nope!

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+  children: PropTypes.node.isRequired,
+  closeable: PropTypes.bool,
+  closeIcon: PropTypes.string,
+  content: PropTypes.oneOfType([
+    PropTypes.node,
+    PropTypes.string,
+    PropTypes.element,
+  ]),
+  direction: PropTypes.oneOf(directions),
+  hover: PropTypes.bool,
```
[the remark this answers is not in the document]
mdespuits: Yes

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+  onClose: PropTypes.func,
+  open: PropTypes.bool,
+  text: PropTypes.string,
+  tooltipOpts: PropTypes.shape({
+    className: PropTypes.string,
+    style: PropTypes.object,
+  }),
+  width: PropTypes.string,
```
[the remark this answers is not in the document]
mdespuits: I agree, but setting a width on more complex content is much harder to do if you require a `style` prop every time. It would probably be better calling this `tooltipWidth` to be clear, but I don't want to force the use of `style` or `className` every time a tooltip is necessary

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+    tooltipClassName,
+    {
+      [styles.open]: open || hovered,
+    }
+  );
+
+  const offsets = useMemo(() => {
+    // Properly center tooltips on the left and right to the center
+    // of the actual element we are attaching the tooltip.
+    if (direction === 'left' || direction === 'right') {
```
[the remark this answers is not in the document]
mdespuits: Ah, I finally got it! Took me a while....

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,149 @@
...
+          {text || tooltipContent}
+        </div>
+      </div>
+    </div>
+  );
+
+  return (
+    <div className={styles.wrapper} {...rest}>
+      {TooltipContent}
+      <div
```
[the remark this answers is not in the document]
mdespuits: Well, now that I'm also trying to incorporate a11y into it, dropping the div is a bit harder with the need to include `aria-describedby`

src/components/Tooltip/index.js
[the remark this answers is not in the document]
pixelbandito: 👍

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,202 @@
...
+  content: tooltipContent,
+  direction,
+  isOpen,
+  showOnHover,
+  tooltipProps: tooltipOptsProp,
+  tooltipWidth,
+  ...rest
+}) => {
+  const [tooltipId] = useState(
+    `tooltip-${Date.now() * Math.floor(100 * Math.random())}`
```
pixelbandito, line 30: Curiosity: How'd you land on this ID generator? Was random not random enough in practice?

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,202 @@
...
+      >
+        {children}
+      </div>
+    </div>
+  );
+};
+
+Tooltip.propTypes = {
+  children: PropTypes.node.isRequired,
+  closeable: PropTypes.func,
```
pixelbandito, line 150: How about the naming convention `onClose`?

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,202 @@
...
+        setOpen(false);
+      }}
+      {...props}
+    >
+      {children({ open, close, toggle })}
+    </Tooltip>
+  );
+};
+
+EasyRichTooltip.propTypes = {
```
pixelbandito, line 199: Are you planning for a non-rich version? If not, would `EasyTooltip` work?

src/components/Tooltip/index.js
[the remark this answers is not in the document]
pixelbandito: I notice you didn't do this, but I'm ok with it for now. We use this pattern, with its flaws, plenty of places in the app now.

src/components/Tooltip/index.js
mdespuits: Well, I wanted to avoid just using numbers. And since everything could get rendered almost at the time time, I wanted to avoid any accessibility clashes. I figured multiplying the date by a random number would be definitively random.

src/components/Tooltip/index.js
mdespuits: Ah, right. I had that before. Thanks for reminding me of switching it back

src/components/Tooltip/index.js
mdespuits: EasyTooltip would be just for text, and just with text can be done just using `Tooltip` by itself. All of the other props are _mostly_ to make the rich content and interaction possible

src/components/Tooltip/index.js
mdespuits: Well, I wanted to avoid just using numbers. And since everything could get rendered almost at the time time, I wanted to avoid any accessibility clashes. I figured multiplying the date by a random number would be definitively random.

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,202 @@
...
+
+  const closeFunc = useCallback(closeable || function() {}, []);
+
+  const handleSetHover = value => () => {
+    if (showOnHover) {
+      setHovered(value);
+    }
+  };
+
+  const handleEscape = e => {
```
sebastianvera, line 45: Since this is only being used within the following `useEffect` why not moving it to the if block inside?
```typescript
  useEffect(() => {
    if (closeOnEscape) {
      const handleEscape = e => {
        if (closeable && e.key === 'Escape') {
          closeable();
        } 
      };

      document.addEventListener('keydown', handleEscape);
      return () => {
        document.removeEventListener('keydown', handleEscape);
      };
    }

    return () => {};
  }, []);
```

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,202 @@
...
+
+      return () => {
+        document.removeEventListener('keydown', handleEscape);
+      };
+    }
+
+    return () => {};
+  }, []);
+
+  useEffect(() => {
```
sebastianvera, line 63: should we use `useLayoutEffect` instead, so we set the height before the browser's paint step? I've found that by using `useLayoutEffect`, for these cases where we need the content's width/height, we avoid flickering on the screen (we might need it if the tooltip is opened when rendering)

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,202 @@
...
+}) => {
+  const [tooltipId] = useState(
+    `tooltip-${Date.now() * Math.floor(100 * Math.random())}`
+  );
+  const tooltipRef = useRef(null);
+
+  const [hovered, setHovered] = useState(false);
+  const [tooltipHeight, setTooltipHeight] = useState(0);
+
+  const closeFunc = useCallback(closeable || function() {}, []);
```
sebastianvera: I think we don't need `useCallback` here, this is:
1. not being used as a dependency for a hook
2. not being pass as a prop down the tree, it's just being used in the button on line 119, so having a different reference no every render will not compromise performance or generate re-renders

Let me know if maybe I missed something 🙂

src/components/Tooltip/story.js
```diff
@@ -0,0 +1,251 @@
...
+          isOpen={boolean('Open', false, 'text')}
+        >
+          <Button>{text('Main Content', 'With Text', 'text')}</Button>
+        </Tooltip>
+      </Element>
+      <Tooltip
+        isOpen={open}
+        showOnHover={boolean('Show on hover', false, 'content')}
+        direction={direction}
+        closeable={toggle}
```
sebastianvera: what do you guys think about having an `onClose` function instead of using this as a toggling function? I feel like by giving this to the user we give them space to use them in the wrong way. Let's say, for example, the user starts with `isOpen` as `true` and then ` <Button onClick={toggle}>With content</Button>` gets clicked, the tooltip will close instead of open, it might sound like a contrived example, and that a user will not do that, but by exposing this API we let them do it.

I'm not proposing to remove this prop, just converting it to a boolean and adding an `onClose` prop.

The EasyRichTooltip would look like this:

```typescript
export const EasyRichTooltip = ({ closeable, onClose = () => {}, children, isOpen = false, ...props }) => {
  const [isOpen, setOpen] = useState(isOpen);

  const open = () => setOpen(true);
  const close = () => { 
    setOpen(false);
    onClose();
  };
  const toggle = () => {
    if (isOpen) {
      close();
    } else. {
      open();
    }
  };

  return (
    <Tooltip
      {...props}
      closeable={closeable}
      onClose={onClose}
      isOpen={isOpen} // I moved it after the spread to avoid overwriting it
    >
      {children({ open, close, toggle })}
    </Tooltip>
  );
};
```

If we decide not to go with this approach I think we should proxy the `isOpen` proxy and also move the props' spread to the top.

cc: @mdespuits @pixelbandito

src/components/Tooltip/style.css
```diff
@@ -0,0 +1,227 @@
...
+  word-wrap: none;
+  text-align: center;
+}
+
+.tooltipClose {
+  background-color: transparent;
+  border: none;
+  cursor: pointer;
+  color: var(--rvr-gray-lite-2);
+  margin: 0;
```
sebastianvera, line 76: should this be removed?

src/components/Tooltip/style.css
```diff
@@ -0,0 +1,227 @@
...
+  margin: 5px;
+  padding: 0;
+  position: absolute;
+  right: 5px;
+  top: 5px;
+  z-index: 51;
```
sebastianvera, line 82: I'm wondering why this is 51, does it have any special meaning?

src/index.js
```diff
@@ -29,3 +29,4 @@ export { default as Collapse } from './components/Collapse';
 export { default as ExpansionPanel } from './components/ExpansionPanel';
 export { default as Accordion } from './components/Accordion';
 export { default as Avatar } from './components/Avatar';
+export { default as Tooltip, UncontrolledTooltip } from './components/Tooltip';
```
sebastianvera, line 32: Wrong component name here, I think I prefer `UncontrolledTooltip` over `EasyRichTooltip`, but I'm ok with either 🚀

src/components/Tooltip/index.js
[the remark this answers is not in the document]
sebastianvera: I just commented about that, I should have refreshed the PR but I was in the middle of the review 😱

src/components/Tooltip/index.js
sebastianvera: I'll add the same comment on the current code since this got outdated

src/components/Tooltip/index.js
```diff
@@ -0,0 +1,202 @@
...
+}) => {
+  const [tooltipId] = useState(
+    `tooltip-${Date.now() * Math.floor(100 * Math.random())}`
+  );
+  const tooltipRef = useRef(null);
+
+  const [hovered, setHovered] = useState(false);
+  const [tooltipHeight, setTooltipHeight] = useState(0);
+
+  const closeFunc = useCallback(onClose || function() {}, []);
```
sebastianvera, line 37: I think we don't need `useCallback` here, this is:
1. not being used as a dependency for a hook
2. not being pass as a prop down the tree, it's just being used in the button on line 119, so having a different reference no every render will not compromise performance or generate re-renders

Let me know if maybe I missed something 🙂

src/index.js
mdespuits: We have the `Easy*` prefixed used in other components. It's simply a pattern we've adopted for Rover. Not attached to it, but worth considering

mdespuits: And yes, I'll fix the export name :)

src/components/Tooltip/index.js
mdespuits: Good point. 👍

src/components/Tooltip/index.js
mdespuits: Oh, that's actually something I hadn't considered. `useLayoutEffect` is still unknown to me, so time to dive in!

src/components/Tooltip/story.js
mdespuits: I've since changed this to `onClose`. I figured since `closeable` will _always_ require an `onClose` behavior (practically speaking) it made sense to use the existence of a function passed into `onClose` as enough to enable that behavior.

I'm not absolutely married to it, but it does reduce the property surface area of the tooltip a bit.

src/components/Tooltip/style.css
mdespuits: Maybe? I'd have to look at my CSS again (since I haven't looked at it in a week)

src/components/Tooltip/style.css
mdespuits: This was just to make sure that the tooltip content, the white arrow, and the arrow shadow had correct z-index relative to each other.

This is a bit of a magical number. I should put it with the other tooltip CSS custom properties

src/components/Tooltip/index.js
mdespuits: I generally try to avoid `useCallback` myself. I think I added this thinking "I don't want to recreate an empty function endlessly". But I could just use an empty function `const` outside of the component or just use `lodash/noop`. Either way, I can drop the overhead of `useCallback`

pixelbandito: The only place you're passing `closeFunc`, it's already wrapped in a condition for `onClose` being truthy.
Just drop the `closeFunc` entirely, and you should be fine.

src/components/Tooltip/style.css
mdespuits: Ah, I didn't see the line right after the comment. Thanks!

PR #158, opened 2020-04-08
Add Callout to exports

pixelbandito:
Callout exists, has a story and tests, but wasn't being exported from the package.

Along the way, I found a couple of places importing lodash utilities in a not-tree-shakeable way, so this PR also reduces the library size from around 1.1 MB to 600 KB minified.

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -21,6 +21,7 @@ import {
   Collapse,
   ExpansionPanel,
   Tooltip,
+  Callout,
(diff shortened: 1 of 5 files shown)


mdespuits:
> Along the way, I found a couple of places importing lodash utilities in a not-tree-shakeable way, so this PR also reduces the library size from around 1.1 MB to 600 KB minified.

## 🎉 🎉 🎉 🎉 🎉 🎉 🎉

PR #159, opened 2020-04-09
TypeScript!

mdespuits:
Entirely optional for now, but worth adding since we are using typescript in our main application as well.


tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+    "outDir": "build",
+    "sourceMap": true /* Generates corresponding '.map' file. */,
+    "removeComments": true /* Do not emit comments to output. */,
+    "noEmit": true /* Do not emit outputs. */,
+    "strict": true /* Enable all strict type-checking options. */,
+    "noImplicitAny": false /* Raise error on expressions and declarations with an implied 'any' type. */,
+    "moduleResolution": "node" /* Specify module resolution strategy: 'node' (Node.js) or 'classic' (TypeScript pre-1.6). */,
+    "baseUrl": "./" /* Base directory to resolve non-absolute module names. */,
+    "allowSyntheticDefaultImports": true /* Allow default imports from modules with no default export. This does not affect code emit, just typechecking. */,
+    "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */
```
dkordik, line 18: huge fan of all these comments!

mdespuits: Just came from `tsc --init`. Can't take credit. 😄

dkordik: ha, nice. well I'm still a fan :D

sebastianvera:
Not sure if this is related to the @types problem https://rollupjs.org/guide/en/#installation

mdespuits:
@sebastianvera No, I don't believe so. The error with the `@types` happens when just running `tsc`, so there is something else that is an underlying issue. We will have to revisit the issue in the future if we need to change the `types: []` config

sebastianvera:
> @sebastianvera No, I don't believe so. The error with the `@types` happens when just running `tsc`, so there is something else that is an underlying issue. We will have to revisit this in the future if we need to change the `types: []` config

Got it, I was curious so I played a little bit with our dependencies and it seems this is related  https://github.com/DefinitelyTyped/DefinitelyTyped/issues/44019 

We can definitely revisit this later

PR #162, opened 2020-04-21
Deploy GH Pages Workflow

mdespuits:
Uses a pre-built Github Workflow action to deploy our GitHub pages documentation.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -28,3 +28,4 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./docs
+          commit_message: "Storybook 🚀: Triggered from ${{ github.sha }}"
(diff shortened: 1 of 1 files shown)


pixelbandito:
If I understand it right, after this change pushing to master will auto-deploy the storybook documentation site.

Since the gh-page is the API documentation for the npm package, would it make sense to only/always update it when the npm package is published? Is it possible to trigger a git workflow based on pushing a new tag that matches a pattern, like `v[0-9]` or `v*`?

pixelbandito:
It would be cool to auto-incrementing the version and npm publish on every push to master by default, but then we'd need to select major/minor/patch on every push to master.

(There was an npm package we looked at that did this, iirc.)

mdespuits:
@pixelbandito Both of those suggestions are fantastic. And yes, they are both attainable. I figured until we could put that setup together, it would be worth having the documentation updates automatically and we could drop the need for the dev dependency altogether.

In any case, I'd like to get this in for now and revisit that flow on Monday now that we have a CI/CD pipeline we can use.

PR #164, opened 2020-04-24
Dropdown circular dependency

mdespuits:
Not a major breaking issue, but definitely an annoyance when building:

```
$ rollup -c

src/index.ts → dist/index.js, dist/index.es.js...
(!) Unresolved dependencies
https://github.com/rollup/rollup/wiki/Troubleshooting#treating-module-as-external-dependency
stream (imported by node_modules/styled-components/dist/styled-components.esm.js)
(!) Circular dependency: src/components/Dropdown/index.js -> src/components/Dropdown/Menu/index.js -> src/components/Dropdown/index.js
created dist/index.js, dist/index.es.js in 5.6s
✨  Done in 7.68s.
```


package.json
```diff
@@ -48,8 +48,10 @@
     "test": "cross-env CI=1 react-scripts test --env=jsdom",
     "test:watch": "react-scripts test --env=jsdom",
     "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
-    "build": "rollup -c",
-    "start": "rollup -c -w",
+    "clean": "rm -rf dist",
```
pixelbandito, line 51: Just curious - was this necessary to get the build to work, or just an enhancement on the side?

mdespuits: Just an enhancement on the side

sebastianvera: Should we use rimraf for this? so we enable non Unix contributors? 🙂

PR #165, opened 2020-04-24
Simplify Rover CSS module scope name

mdespuits:
This simplifies the CSS class name generation considerably. Instead of `.styles__Avatar-{hash}` format, it now is simply `.rvr-{hash}`. Not only does this clarify that this is a Rover class, but it shortens the classname itself and shrinks the build output a bit

### DIFF
--- a/example/yarn.lock
+++ b/example/yarn.lock
@@ -5041,7 +5041,7 @@ lodash.uniq@^4.5.0:
   resolved "https://trendkite.jfrog.io/trendkite/api/npm/libs-frontend/lodash.uniq/-/lodash.uniq-4.5.0.tgz#d0225373aeb652adc1bc82e4945339a842754773"
   integrity sha1-0CJTc662Uq3BvILklFM5qEJ1R3M=
 
-"lodash@>=3.5 <5", lodash@^4.15.0, lodash@^4.17.11, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.2, lodash@^4.17.3, lodash@^4.17.4, lodash@^4.3.0:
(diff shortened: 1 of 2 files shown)


example/yarn.lock
```diff
@@ -5041,7 +5041,7 @@ lodash.uniq@^4.5.0:
   resolved "https://trendkite.jfrog.io/trendkite/api/npm/libs-frontend/lodash.uniq/-/lodash.uniq-4.5.0.tgz#d0225373aeb652adc1bc82e4945339a842754773"
   integrity sha1-0CJTc662Uq3BvILklFM5qEJ1R3M=
 
-"lodash@>=3.5 <5", lodash@^4.15.0, lodash@^4.17.11, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.2, lodash@^4.17.3, lodash@^4.17.4, lodash@^4.3.0:
+"lodash@>=3.5 <5", lodash@^4.15.0, lodash@^4.17.11, lodash@^4.17.13, lodash@^4.17.14, lodash@^4.17.15, lodash@^4.17.2, lodash@^4.17.3, lodash@^4.17.4, lodash@^4.3.0:
```
pixelbandito, line 5044: Guess we missed a `yarn install` on a previous deploy?

mdespuits: Maybe? This is the `example/yarn.lock` which doesn't get updated as often

dkordik:
> This simplifies the CSS class name generation considerably. Instead of `.styles__Avatar-{hash}` format, it now is simply `.rvr-{hash}`. Not only does this clarify that this is a Rover class, but it shortens the classname itself and shrinks the build output a bit

is there a debugging cost to this? inspecting an element no longer tells you what component is applying a style?

mdespuits:
@dkordik There may be, but 99% of the time active development on Rover will be using the webpack config which includes the name. For the rollup production build it makes sense to use smaller classnames and rely on Component name and CSS sourcemaps to keep the dependencies smaller

PR #173, opened 2020-05-14
TypeScript, microbundle, and external CSS

### LINKED ISSUE #171: Look into generating external CSS file
Right now, all styles for RoverUI are dependent on `styleInject` to include styles for the components we use. This is not ideal as the order of injection is impossible to control. Ideally, we would provide a way to inject RoverUI's CSS earlier in a build to ensure that any consumer's styles take precedence.

mdespuits:
## Three major changes

1. Adds full TypeScript support so any existing components can be converted and new components can use it.
2. Drops Rollup in favor of simpler `microbundle` (`microbundle-crl`, technically) to bundle dependencies.
3. All CSS output automatically compiled to external CSS file.

The final CSS change is what will make this a **major breaking change** and will require a major version bump.

Fixes #171


package.json
```diff
@@ -70,63 +69,68 @@
...
     "react-dom": "^16.9.0",
     "react-scripts": "^1.1.4",
-    "rollup": "^0.64.1",
-    "rollup-plugin-babel": "^3.0.7",
-    "rollup-plugin-commonjs": "^9.1.3",
-    "rollup-plugin-node-resolve": "^3.3.0",
-    "rollup-plugin-peer-deps-external": "^2.2.0",
-    "rollup-plugin-postcss": "^1.6.2",
-    "rollup-plugin-url": "^1.4.0",
+    "rimraf": "^3.0.2",
```
sebastianvera, line 129: I think we can get rid of this one now

tsconfig.json
```diff
@@ -1,23 +1,30 @@
...
     "forceConsistentCasingInFileNames": true /* Disallow inconsistently-cased references to the same file. */,
-    "types": [] /* Don't try to include @types packages right now. Seems to break `tsc` call */
+    "types": [] /* Don't try to include @types packages right now. Seems to break `tsc` call */,
+    "declaration": true,
+    "esModuleInterop": true,
+    "noEmit": true,
+    "noImplicitReturns": true,
+    "noImplicitThis": true,
+    "noImplicitAny": false,
+    "skipLibCheck": true,
```
sebastianvera, line 22: is it safe to enable this? 
>  Skip type checking of all declaration files (*.d.ts).

tsconfig.json
```diff
@@ -1,23 +1,30 @@
...
+    "strictNullChecks": true,
+    "suppressImplicitAnyIndexErrors": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true
   },
-  "include": ["src/**/*"],
-  "exclude": ["node_modules"]
+  "include": ["src"],
+  "exclude": ["node_modules", "dist", "example", "node_modules/@types"]
```
sebastianvera, line 29: isn't specifying `"types": []` enough? or do we still need to exclude `node_modules/@types`?

mdespuits: The `"types": []` was to fix an issue with a particular version of node and TS having issues

pixelbandito:
Final thing: I assume you verified storybook and the examples/index test page.

Did you by chance try using `yalc` / `yarn link` and testing with the real front-end app?

I'm not sure how necessary it is, but it would be nice to check.

mdespuits:
> Final thing: I assume you verified storybook and the examples/index test page.
> 
> Did you by chance try using yalc / yarn link and testing with the real front-end app?
> 
> I'm not sure how necessary it is, but it would be nice to check.

Yes to all three and all three are looking great! I have a follow up branch that makes updates to the example App to include Tailwind and some nicer wrappings.

PR #174, opened 2020-05-15
Post-microbundle updates

mdespuits:
## Adds CONTRIBUTING.md

Github's official file used for contributing guidelines. Simplifies the main README quite a lot, too

## Updates the example App

* Adds Tailwind from CDN for nicer looking examples
* Fixes bugs found through TypeScript as console error logs

Kapture 2020-05-15 at 11 51 27

## Updates the `hygen` generated code

Finding component files in Rover has always been a bit difficult due to everything being called `index.js` or `style.css`. For future components, I'm proposing switching to use `index.ts` for simple exporting and `Component.tsx` and `Component.module.css` for easier to file finding.


.github/workflows/gh-pages.yml
```diff
@@ -21,11 +20,10 @@ jobs:
       - name: Install and Build 🔧
         run: |
           yarn install
-          yarn build-storybook
+          yarn storybook:build
```
mdespuits, line 23: This was the real problem. `build-storybook` does not build to docs by default

.eslintrc
```diff
@@ -1,6 +1,5 @@
 {
   "root": true,
-  "parser": "@typescript-eslint/parser",
```
sebastianvera, line 3: just curious about this change, is it not necessary anymore?

_templates/component/new/Component.module.css.t
```diff
@@ -0,0 +1,7 @@
+---
+to: src/components/<%= h.changeCase.pascal(name) %>/<%= h.changeCase.pascal(name) %>.css
```
sebastianvera, line 2: is the `module` extension missing here?

mdespuits: I wasn't entirely sure I wanted to add that much "noise" to the file if it already is setup to work without it. Worth having a discussion, though

.eslintrc
mdespuits: I don't believe so. I was actually getting errors when adding new .tsx files and my Googling indicated this could be part of the problem. Removing it made all the errors vanish, but I'm willing to try it again as well.

_templates/component/new/Component.module.css.t
sebastianvera: it's just a convention for when using CSS modules, files without the `.module.` extension are considered global styles

mdespuits: Yeah, that is true. I'm happy to make the switch.

CODE_GUIDELINES.md
```diff
@@ -12,12 +12,14 @@ Instead, write components with generic objects in mind, like people, media, acti
...
-```
+```sh
+$ tree src/components/Modal
 Modal/
-|-- index.js   # Core component file, imports style.css
-|-- style.css  # Styles, as CSS modules
-|-- test.js    # Used with an npm command
-`-- story.js  # Used by storybook, separate npm command
+|-- index.ts          # Re-exports from Modal.tsx
+|-- Modal.tsx         # Core component file, imports Modal.css
```
pixelbandito, line 19: Bikeshedding:
I know I read a couple of places that newer recs were to use `.js` and not `.jsx`, even with JSX in the files.
The argument boiled down to "everything is transpiled anyway, and not having to rename files is better ergonomics".

Would that make sense with TS files too (skip the trailing "x"?)
I don't mind much either way, but I have a slight preference for using the same extension for all.

CONTRIBUTING.md
```diff
@@ -0,0 +1,75 @@
+# Contributing to RoverUI
```
pixelbandito, line 1: Good file addition, thanks very much!

CODE_GUIDELINES.md
mdespuits: I have not looked into this much, but when I leave off the `s` for TS components, TS definitely complains loudly. With the `x`, it happily parses it.

.eslintrc
mdespuits: So, it may actually just be an issue with VS code. The error I was getting _only_ occurred with new component files. When I reloaded VS code, it seemed to pick up the correct parser: https://github.com/typescript-eslint/typescript-eslint/issues/864

Now that I know the issue, I can work around it. Re-adding parser

CODE_GUIDELINES.md
mdespuits: @pixelbandito Ah, here's an explanation for this: https://www.typescriptlang.org/docs/handbook/jsx.html#basic-usage. For explicit TS, we need to use the `.tsx` extension to enable JSX for TypeScript components

PR #178, opened 2020-05-22
TypeScript: Tests, CSS modules, and dependency updates

mdespuits:
As it turns out, trying to convert existing components and their tests, etc to TypeScript is currently tricky with our outdated dependencies. One of the primary offenders was `react-scripts` itself. We were running on a 2 year old version. It was worth updating to the latest anyway since it has much better TypeScript support. This meant several things:

* Upgrading `react-scripts` changed the matcher for Jest tests. Needed to be `*.(spec|test).(ts|js|jsx|tsx)` or something along those lines
* Completely removing the custom css-loader config for Storybook as the `react-scripts` webpack default configuration works _almost_ without customization. I decided to just rename `style.css` to `{Component}.module.css` in keeping with normal conventions for CSS modules.
* Converted the `Responsive.Container` to TypeScript to test the problems with the current `beta.2` installation in our main app (something to do with auto-generated types with PropTypes)
* Went ahead and completely converted the `Pill` component to the updated format (minus the `story.js` file):
  ```sh
  ❯ tree src/components/Pill
  src/components/Pill
  ├── Addon.tsx
  ├── Pill.module.css
  ├── Pill.test.tsx
  ├── Pill.tsx
  ├── README.md
  ├── index.tsx
  └── story.js
  ```


src/components/Accordion/Accordion.test.js
```diff
@@ -39,7 +39,7 @@ describe('Accordion', () => {
         </ExpansionPanel>
       </Accordion>
     );
-    wrapper.children().forEach(panel => {
+    wrapper.children().forEach((panel) => {
```
pixelbandito, line 42: Could we change this lint rule?
Personal preference: I think skipping parens on single argument arrow functions is nice and clean.

package.json
```diff
@@ -83,34 +83,37 @@
...
+    "babel-eslint": "^10.0.3",
     "babel-loader": "^8.1.0",
     "babel-plugin-external-helpers": "^6.22.0",
     "babel-plugin-transform-es2017-object-entries": "^0.0.5",
     "cross-env": "^7.0.2",
     "cssnano": "^4.1.10",
     "enzyme": "^3.9.0",
     "enzyme-adapter-react-16": "^1.13.0",
-    "eslint": "^7.0.0",
+    "eslint": "^6.8.0",
```
pixelbandito, line 102: I assume this downgrade was connected to the babel-eslint one?

package.json
```diff
@@ -83,34 +83,37 @@
...
     "@types/jest": "^25.1.4",
+    "@types/node": "^14.0.4",
     "@types/react": "^16.9.27",
-    "@typescript-eslint/eslint-plugin": "^2.33.0",
+    "@types/react-dom": "^16.9.8",
+    "@typescript-eslint/eslint-plugin": "^2.26.0",
     "@typescript-eslint/parser": "^2.26.0",
     "auto-changelog": "^2.0.0",
-    "babel-eslint": "^10.1.0",
+    "babel-eslint": "^10.0.3",
```
pixelbandito, line 94: Why the downgrade?

src/components/Avatar/Addon.js
```diff
@@ -2,7 +2,7 @@ import React, { useMemo } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames';
 
-import style from './style.css';
+import style from './Avatar.module.css';
```
pixelbandito, line 5: Nice

src/components/Avatar/index.js
```diff
@@ -17,7 +17,7 @@ function defaultImageFetcher({ onLoad, onError, src }) {
   const img = new Image();
   img.referrerPolicy = 'no-referrer';
   img.crossOrigin = 'Anonymous';
-  img.onload = function() {
+  img.onload = function () {
```
pixelbandito, line 20: I don't really care about `function()` vs. `function ()`, and it seems to be purely stylistic, not avoiding possible errors.

In that case, the question is: why the change? Are we opting into a new version of recommendations as a byproduct of this PR?

src/components/Collapse/index.js
```diff
@@ -36,17 +36,17 @@ const Collapse = ({
...
     },
     [onExit, setHeight]
   );
 
   const handleExiting = useCallback(
-    node => {
-      /* eslint-disable no-unused-expressions */
-      node.offsetHeight; // getting this variable triggers a reflow
+    (node) => {
+      // eslint-disable-next-line @typescript-eslint/no-unused-expressions
```
pixelbandito, line 48: Nice, thank you.

src/components/Avatar/index.js
pixelbandito: I guess my rule of thumb is: shorter is better, if it's not a legibility or avoiding bug issue.
But if this is blessed by the folks who do eslint recommendation presets, I give them a little extra benefit of the doubt.

src/components/Callout/index.js
```diff
@@ -35,7 +35,7 @@ const Callout = ({
       {children}
       {onCancel && isFunction(onCancel) && (
         <div className={styles.cancel}>
-          <button className={styles.cancelButton}>
+          <button type="button" className={styles.cancelButton}>
```
pixelbandito, line 38: Nice

src/components/Accordion/Accordion.test.js
mdespuits: Yes, we can. I don't mind.

package.json
mdespuits: No, this was actually connected to `react-scripts`. Kind of disappointing, but it's dependency is on `^6.6.0` 😢

package.json
mdespuits: Again, probably due to `react-scripts`

src/components/Avatar/index.js
mdespuits: Ah! This is probably a change in eslint defaults? I don't care for it either

mdespuits: Turns out Prettier 2.0 forces spaces between `function` and the `()`s. Reverted to 1.19

package.json
mdespuits: Nope, this was something else. Reverting!

PR #184, opened 2020-05-24
Badge => TypeScript

### LINKED ISSUE #182: Convert Badge to TypeScript
Fairly small footprint and a fairly simple component.

mdespuits:
## EXAMPLE. DO NOT MERGE YET.

This is an example of converting a JS component to TS. Step I took

1. Move `index.js` to `Badge.tsx` and create `index.ts` to re-export from the new Badge TS file.
2. Run tests.
3. Fix any immediate TS errors
4. Commit
5. Use `React.FC` and setup correct Badge props and resolve conflicts with `PropTypes` auto-typing.

I only made two small additions:

* Deprecate `round` to prefer `isRound`. Open to reverting this.
* Add typing to possible variants for useful autocomplete when using with consuming apps.

Fixes #182.


src/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,80 @@
...
+};
+
+Badge.defaultProps = {
+  className: '',
+  round: null,
+  isRound: null,
+  variant: '',
+};
+
+Badge.propTypes = {
```
pixelbandito, line 73: Can you remove the old propTypes / defaultProps now that there are typescript equivalents?

src/components/Badge/story.js
```diff
@@ -46,10 +46,8 @@ storiesOf('Planets/Badge', module)
   .add(
     'Overview',
     () => {
-      const colorOptions = variants;
-
       const fontSize = select('Font size', fontSizes, 'sm');
-      const variant = select('Variant', colorOptions, '');
+      const variant = select('Variant', variants, '');
```
pixelbandito, line 50: Small thing, but I think it makes more sense to label the knobs with the precise prop name, so doc readers don't have to translate to camelcase in their heads.

mdespuits: Ah, fair enough! Hadn't considered that. I can update it

PR #185, opened 2020-05-26
Only include declarations for TS files

mdespuits:
This actually makes a lot of sense to include declaration files from TypeScript files only instead of any JS files that happen to have "typed" support. This solves all of the issues that we were seeing when trying to setup 2.x.


tsconfig.json
```diff
@@ -27,6 +27,13 @@
     "resolveJsonModule": true,
     "isolatedModules": true
   },
-  "include": ["src"],
-  "exclude": ["node_modules", "dist", "example"]
+  "include": ["src/**/*.ts", "src/**/*.tsx"],
```
pixelbandito, line 30: Should we also include `"src/*.ts", "src/*.tsc",` ?

mdespuits: We don't want to include everything, per-se. Having `setupTests.d.ts` wouldn't be particularly useful

PR #186, opened 2020-05-27
Use Babel plugins to fix Fragments issue

mdespuits:
Apparently, this has been a known issue for a while and the `Fragment` keyword in the final build was not correctly setup. Adding two Babel plugins not only assure backwards-compatibility but also ensure that Fragments will work as expected until these microbundle and TS bugs are resolved and released.

https://github.com/developit/microbundle/pull/623

A known issue in TypeScript:
https://github.com/microsoft/TypeScript/issues/20469

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -12,7 +12,7 @@
     [
       "@babel/plugin-transform-react-jsx",
       {
-        "pragmaFrag": "React.Fragment" // default is React.Fragment
+        "pragmaFrag": "React.Fragment"
       }
     ],
     [
--- a/.babelrc
+++ b/.babelrc
@@ -7,5 +7,19 @@
       }
     ],
     "@babel/preset-react"
+  ],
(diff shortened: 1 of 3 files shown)


.babelrc
```diff
@@ -7,5 +7,19 @@
       }
     ],
     "@babel/preset-react"
+  ],
+  "plugins": [
+    [
+      "@babel/plugin-transform-react-jsx",
+      {
+        "pragmaFrag": "React.Fragment" // default is React.Fragment
```
pixelbandito, line 15: If this is the default, can you omit the options block?

mdespuits: No, actually that's wrong. I tried the default and it just leaves the bare `Fragment`, which is what was breaking in the first place.

pixelbandito: Maybe delete or clarify the comment, then; but very minor. 👍

PR #204, opened 2020-06-01
Add basic Input component wrapper. 

### LINKED ISSUE #202: New Component Idea: Input (text first)
Relates #177 

**Is your feature request related to a problem? Please describe.**
Bare wrapper around HTML inputs for now, with Cision-approved styles.

**Describe the component you'd like**
I'm planning on making an `Input` component that generally passes props down.

**How would you use this component?**
This component will be extended for other types, like time.

```jsx
<Input
  value="text"
  onChange={handleChange}
/>
```

pixelbandito:
I'll need to do this separately against master (v2 beta)

Fixes #202

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "name": "@cision/rover-ui",
-  "version": "1.5.5",
+  "version": "1.5.0",
   "publishConfig": {
-    "tag": "v1.5.5"
+    "tag": "v1.5.0"
   },
   "description": "UI Component Library",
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "@cision/rover-ui",
-  "version": "1.4.5",
+  "version": "1.5.5",
```
mdespuits, line 3: Probably should be `1.5.0`

pixelbandito: 🤦

PR #206, opened 2020-06-02
Add basic `Input` component to master / v2.x

### LINKED ISSUE #202: New Component Idea: Input (text first)

pixelbandito:
Fixes #202

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cision/rover-ui",
-  "version": "2.0.0-beta.6",
+  "version": "2.0.0-beta.5",
   "publishConfig": {
     "tag": "next"
   },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cision/rover-ui",
-  "version": "2.0.0-beta.5",
+  "version": "2.0.0-beta.6",
   "publishConfig": {
     "tag": "next"
   },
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cision/rover-ui",
-  "version": "2.0.0-beta.5",
+  "version": "2.0.0-beta.6",
```
mdespuits, line 3: We should establish a the pattern for when to bump versions. I'm fine with this here, but I think a commit of it's own is warranted (especially to keep the CHANGELOG consistent and properly up-to-date)

pixelbandito: Oh, snap, I forgot about the changelog.

I don't mind reverting that mit and making a separate PR for the version bump. Is there a special incantation I need to use to make the changelog nice?

mdespuits: It's just `yarn update-changelog`. My current workflow for releasing new versions has been.

* `git checkout master`
* Bump version in package.json
* `yarn update-changelog`
* `git commit -m "v{version}"`
* `git push`
* `yarn release --tag next`

PR #208, opened 2020-06-06
Input more docs and cleanup

mdespuits:
* Adds `readOnly` and `onChange` defaults to `Input`
* Moves `.stylelintrc` to `stylelint.config.js`. Seems to actually work correctly?
* More style cleanup based on Figma docs to avoid style conflicts
* Convert Input story file to TS


src/components/Input/Input.module.css
```diff
@@ -4,29 +4,17 @@
   border: var(--rvr-border-width-md) solid var(--rvr-gray-60);
   color: var(--rvr-gray-90);
   font-family: var(--rvr-base-font-family);
-  font-size: var(--rvr-font-size-md);
   line-height: var(--rvr-line-height-lg);
+  outline: none;
```
mdespuits, line 8: Since we provide our own `focus` styles, we don't want the browser's default in this case.

src/components/Input/Input.module.css
```diff
@@ -4,29 +4,17 @@
...
   padding: var(--rvr-space-bordered-sm);
-  transition:
-    100ms linear background-color,
-    100ms linear border-color,
-    100ms linear box-shadow,
-    100ms linear color;
+  transition: 100ms linear background-color, 100ms linear border-color,
```
mdespuits, line 10: Stylelint is responsible for this. I believe this was something that should have been caught before, but was not being caught/fixed properly with the `.stylelintrc` file.

src/components/Input/Input.tsx
```diff
@@ -4,14 +4,17 @@ import { ClassValue } from 'classnames/types';
 
 import styles from './Input.module.css';
 
-interface InputProps {
+interface InputProps
+  extends Omit<React.InputHTMLAttributes<HTMLInputElement>, 'className'> {
   className?: ClassValue;
```
mdespuits, line 9: Unrelated: Not actually sure we should do this the more I think about it. Probably should stick with the default since  we don't necessarily need to expect `className` to be anything other than `string`. I think that should be up to the consumer how to generate the class string here. Thoughts? (@pixelbandito @dkordik)

pixelbandito: Can you elaborate on what the "right" way to do it is?

Are you just saying stick with `className?: string;` I'd be down with that.

I have no interest in supporting consumers passing ClassMaps into these components, they can make their own darn string.

src/components/Input/Input.module.css
```diff
@@ -49,14 +37,28 @@
...
+.Input.disabled:hover {
+  background-color: var(--rvr-white);
+}
+
+.Input:disabled::placeholder,
+.Input.disabled::placeholder {
+  color: var(--rvr-gray-20);
+}
+
+.Input:not(:focus):not(:disabled):hover {
```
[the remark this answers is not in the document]
pixelbandito: I'd suggest not change it with this PR. 
I'm still actively developing `Input` variants, like the time picker, and I'll be talking to design and tweaking over the next few weeks.

src/components/Input/Input.module.css
mdespuits: Maybe this was due to some of the re-arranging I was doing in response to suggestions from `stylelint`. I'll revisit this on Monday

src/components/Input/Input.tsx
mdespuits: If we want to stick with the `className?: string`, then we don't have to do anything at all in this case since that is what is provided as part of `React.InputHTMLElement` and other similar interfaces provided by React.

src/components/Input/Input.module.css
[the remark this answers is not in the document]
mdespuits: I'll revert these changes to avoid any conflicts

src/components/Input/Input.tsx
```diff
@@ -1,17 +1,17 @@
...
-interface InputProps {
-  className?: ClassValue;
+interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> {
   /** Applies disabled appearance without disabling. Helpful for adding tooltips on disabled field clicks. */
   fauxDisabled?: boolean;
 }
 
 const Input: React.FC<InputProps> = ({
   className = '',
+  onChange = () => {},
```
pixelbandito, line 13: Do you need to add onChange and readOnly? I think they should be handled fine by passedProps.

mdespuits: This was actually due to some suggestions from React warnings. This was what it said:

> checkPropTypes.js:20 Warning: Failed prop type: You provided a `value` prop to a form field without an `onChange` handler. This will render a read-only field. If the field should be mutable use `defaultValue`. Otherwise, set either `onChange` or `readOnly`.

I didn't know `defaultValue` existed, but it seemed reasonable to provide defaults based on this.

mdespuits: Also, thank you for asking. I completely forgot to leave a comment clarifying the addition

pixelbandito: Looks good, thanks for clarifying!

PR #211, opened 2020-06-08
Grid.Entry => TypeScript

mdespuits:
Related to #199


src/components/Grid/Entry/Entry.tsx
```diff
@@ -3,12 +3,23 @@ import PropTypes from 'prop-types';
 
 import styles from './Entry.module.css';
 
+interface EntryProps extends React.HTMLAttributes<HTMLDivElement> {
+  entryPercentWidth?: number;
```
pixelbandito, line 7: Children is required in the original propTypes, but not in the TS interface

pixelbandito: Also, did we decide we needed to leave the original propTypes and defaultProps, or can they be removed safely now?

mdespuits: I believe we agreed to drop them since we are the primary consumers of the library. Plus, trying to have TypeScript and PropTypes together is tricky in certain situations.

mdespuits: > Children is required in the original propTypes, but not in the TS interface

I missed this. I can add it back using `Omit`

mdespuits: Turns out you don't need to use `Omit` if you are just making a prop required

src/components/Grid/Entry/Entry.tsx
```diff
@@ -1,14 +1,26 @@
-import React from 'react';
+import React, { Fragment } from 'react';
```
mdespuits, line 1: There is currently an issue with TypeScript and Fragments that I cannot track down, but using `<>` is not working nicely: https://github.com/microsoft/TypeScript/pull/35392

PR #212, opened 2020-06-11
Collapse => TypeScript

### LINKED ISSUE #192: Convert Collapse to TypeScript

mdespuits:
Fixes #192.


mdespuits:
@pixelbandito Would it be reasonable to pass any additional props onto the underlying `<div />`

mdespuits:
@sebastianvera before i merge I’m going to pass extra props to the div in the case we want to pass any custom classes (for non-height adjustments)

mdespuits:
@sebastianvera Turns out it isn't really necessary. Any styling of inner content should be done with the children themselves anyways and this can focus entirely on height.

PR #213, opened 2020-06-12
TabMenu => TypeScript

### LINKED ISSUE #194: Convert TabMenu and TabMenu.Item to TypeScript

mdespuits:
Accomplishes two things:

* Converts all `TabMenu` components to TS
* Deprecates setting a plain string of `active` as a way to change styling rather than rely on the existing property.

I also add an `Example` section to the TabMenu storybook

Fixes #194


src/components/TabMenu/Item/Item.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export interface ItemProps extends React.HTMLAttributes<HTMLLIElement> {
+  active?: boolean;
+}
+
+const Item: React.FC<ItemProps> = ({
+  className: customClassName,
+  active = false,
+  children,
+  onClick = () => {},
```
mdespuits, line 14: @pixelbandito @sebastianvera @dkordik As I opened this PR, I noticed that this `onClick` handler is simply dropping any ability to listen to click events on this component. Seems like a bug. Thoughts on removing this?

src/components/TabMenu/Item/Item.tsx
```diff
@@ -0,0 +1,40 @@
...
+  ...props
+}) => {
+  const isChildActive = (child: React.ReactNode) => {
+    if (React.isValidElement(child)) {
+      const activeChild = child.props?.className?.indexOf('active') >= 0;
+      if (activeChild) {
+        console.warn(
```
mdespuits, line 21: This seemed like a reasonable way to add deprecations for now since we don't have many.

pixelbandito: Good call, but it still adds tech debt. How would you feel about looking for instances that use this, squash them, and eliminate the code?

Or just making an issue to track that work.

pixelbandito: On second thought, this may specifically be to cover react router's `Link` components.
If that's the case, this might be a helpful, although slightly weird, feature.

mdespuits: That is actually a very good point. I'll take a closer look at that possibility on Monday. Thanks for catching this!

pixelbandito:
Can you verify that tab content missing `className` doesn't break when this goes out.

src/components/TabMenu/Item/Item.tsx
pixelbandito: I think the only thing this PR is waiting for is the answer to this question.
If we find out it's a NavLink / Link enabler, then the only code change would be switching from a `console.warn` to a comment explaining why the weird code is here.

nvmusoke: @mdespuits did you ever look into this? And can you or @pixelbandito explain what's happening here so I can try to resolve the issue?

src/components/TabMenu/Item/Item.tsx
dkordik: does seem odd that this prop is given a default value if it's not actually used in the component... seems safe to remove as it's unused. (I wonder if the blame would reveal anything interesting- maybe it was formerly used and never cleaned up?)

src/components/TabMenu/Item/Item.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+const Item: React.FC<ItemProps> = ({
+  className: customClassName,
+  active = false,
+  children,
+  onClick = () => {},
+  ...props
+}) => {
+  const className = classNames(styles.Item, customClassName, {
+    [styles.active]: active || children,
```
dkordik, line 18: this will now set the active class if "children" is truthy? will that not give false positives for inactive tabs? (where "children" would contain the node with the name of the inactive tab?)

dkordik: maybe a good first step would be to keep the functionality as is for the TS conversion, while adding some `?.` to that section so it no longer errors when there is no child node with a className prop? 
```javascript
    if (React.isValidElement(child)) {
      return child.props?.className?.indexOf('active') >= 0;
    }
```

PR #217, opened 2020-06-15
Responsive => TypeScript

### LINKED ISSUE #189: Convert all Responsive subcomponents to TypeScript

mdespuits:
I tried to do this a bit more piecemeal, but it seems like there are too many inter-dependencies for that to happen cleanly. Went ahead and converted everything related to Responsive-ness to TS.

Fixes #189


src/components/Responsive/Element/Element.tsx
```diff
@@ -0,0 +1,17 @@
+import React, { ReactElement } from 'react';
+import Context from '../Context';
+
+const Element: React.FC = ({ children, ...passedProps }) => (
```
pixelbandito, line 4: Did it complain when typing children with something like `children?: React.ReactNode` ?

src/components/Responsive/Grid/Grid.tsx
```diff
@@ -1,10 +1,34 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import React, { ReactElement } from 'react';
 
 import UnresponsiveGrid from '../../Grid';
 import Context from '../Context';
 
-const Grid = ({ breakpoints, children, columns, gutter, ...passedProps }) => {
+type GridGutter = number | string;
```
pixelbandito, line 6: Super minor, but I'd usually put an empty newline before the multiline `type GridDefinition ...`

Not a big deal, but I like the consistency. It should be a pretty easy thing to eslint and autofix, if others agree.

src/components/Responsive/Hidden/Hidden.tsx
```diff
@@ -0,0 +1,44 @@
+import React, { useCallback, ReactElement } from 'react';
+import intersection from 'lodash/intersection';
+
+import Context from '../Context';
+
+interface HiddenProps {
+  breakpoints: string[];
```
pixelbandito, line 7: Breakpoints wasn't previously required. the lodash `intersection` method is null / undefined safe, and just returns an empty array.

Small, but I think this merits a fix.

src/components/Responsive/Hidden/Hidden.test.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { mount } from 'enzyme';
 import Context from '../Context';
-import Hidden from './';
+import Hidden from '.';
 
 describe('<Hidden />', () => {
   it('renders without error', () => {
     mount(
-      <Hidden>
+      <Hidden breakpoints={[]}>
```
pixelbandito, line 9: See my comment below - this shouldn't need to be required.
I agree it doesn't make much sense to omit it, but in the spirit of robustness "be conservative in what you do, be liberal in what you accept from others"

pixelbandito: Postel's law, if you're not familiar. I love how it was written about TCP specs, but it's also good ethics.
https://en.wikipedia.org/wiki/Robustness_principle

src/components/Responsive/Visible/Visible.test.tsx
```diff
@@ -6,7 +6,7 @@ import Visible from '.';
 describe('<Visible />', () => {
   it('renders without error', () => {
     mount(
-      <Visible>
+      <Visible breakpoints={[]}>
```
pixelbandito, line 9: See my comment on `Hidden` - see if we can keep `breakpoints` optional?

src/components/Responsive/story.tsx
```diff
@@ -6,12 +6,13 @@ import { object } from '@storybook/addon-knobs';
...
 import ResponsiveElementReadme from './Element/README.md';
-import Responsive from './';
+import Responsive from '.';
+import { defaultBreakpoints } from './Container/Container';
 
 const MyComponent = ({ responsiveContext }) => (
   <div>
     My component&apos;s `responsiveContext` prop:
-    <pre style={{ whiteSpace: 'wrap', wordWrap: 'break-word' }}>
+    <pre className="break-words">
```
pixelbandito, line 15: 😉

src/components/Responsive/Element/Element.tsx
mdespuits: No, `React.FC` handling of the `children` prop is so simple and nice. No complaints at all.

PR #220, opened 2020-06-16
ExpansionPanel => TypeScript

mdespuits:
* `ExpansionPanel`, `ExpansionPanel.Header`, and `ExpansionPanel.Body` all converted to TypeScript
* Updates stories to use some Tailwind by default to help with styling and clarify documentation a bit.

Related to #101


src/components/ExpansionPanel/ExpansionPanel.test.tsx
```diff
@@ -11,7 +12,7 @@ describe('ExpansionPanel', () => {
       </ExpansionPanel>
     );
 
-    expect(wrapper.prop('expanded')).toBe(false);
+    expect(wrapper.find(Collapse).prop('isOpen')).toBe(false);
```
sebastianvera, line 15: here we are testing implementation details, wouldn't it be better if we assert the content is not visible?

src/components/ExpansionPanel/ExpansionPanel.tsx
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  return (
+    <div {...passedProps}>
+      {header &&
+        React.cloneElement(header as ReactElement, {
+          expanded,
+          onClick: handleToggle,
+        })}
+      <Collapse isOpen={expanded ?? false} {...collapseProps}>
```
sebastianvera, line 62: can `expanded` really be `undefined`?

src/components/ExpansionPanel/ExpansionPanel.test.tsx
mdespuits: Probably. I was just trying to get tests to pass rather than trying to pass for this change rather than testing things better. We will have React `testing-library` added soon with some of @pixelbandito's changes, so that will be a set in the right direction for that.

Also, the content is still there in the HTML, but the CSS changes, so we would still be asserting against CSS selectors.

src/components/ExpansionPanel/ExpansionPanel.tsx
mdespuits: Well, TypeScript helps make sure it will be a boolean, but there is a possibility it ends up becoming `null` or `undefined`. This is just to be explicit. But I suppose that `!!expanded` would accomplish the same thing?

PR #222, opened 2020-06-19
Support Date type for value, min, max

pixelbandito:
Builds on the InputTime component, by allowing the user to pass in a ISO date string.

The use case is for pairing the time picker with a date picker in the future for date-time goodness.

1. parses the date value to get the expected time string format and passes it along.
2. for min and max dates: 
  a. discards min/max if the entire current day is valid
  b. disables the input if the entire current day is invalid
  c. parses and passes down the min/max if today is only partially valid.

### DIFF
--- a/src/components/InputTime/AsDate.tsx
+++ b/src/components/InputTime/AsDate.tsx
@@ -1,4 +1,5 @@
 import React, {
+  Fragment,
   MutableRefObject,
   useEffect,
   useImperativeHandle,
@@ -138,7 +139,7 @@ export const AsDate: React.FC<AsStringProps> = ({
   };
 
   return (
-    <>
+    <Fragment>
       <InputTimeAsString
         {...passedProps}
         disabled={invalidMax || invalidMin || disabled}
@@ -154,7 +155,7 @@ export const AsDate: React.FC<AsStringProps> = ({
         style={{ display: 'none' }}
         tabIndex={-1}
       />
-    </>
+    </Fragment>
   );
 };
 
--- a/src/components/InputTime/AsDate.tsx
+++ b/src/components/InputTime/AsDate.tsx
@@ -31,15 +31,24 @@ export const AsDate: React.FC<AsStringProps> = ({
 
   useEffect(() => {
     if (maxDateString && Number.isNaN(new Date(maxDateString).valueOf())) {
-      console.warn('`max` should be an ISO date string');
+      /* eslint-disable-next-line no-console */
+      console.warn(
+        '`max` should be an ISO date string when using `InputTime` with dates'
+      );
     }
 
     if (minDateString && Number.isNaN(new Date(minDateString).valueOf())) {
-      console.warn('`min` should be an ISO date string');
+      /* eslint-disable-next-line no-console */
+      console.warn(
+        '`min` should be an ISO date string when using `InputTime` with dates'
+      );
     }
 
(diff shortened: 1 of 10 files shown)


pixelbandito:
Ah, I need to add tests and some new storybook knobs.

src/components/InputTime/InputTime.tsx
```diff
@@ -25,12 +29,115 @@ Desired features:
...
+    if (maxDate === undefined || maxDate === null) {
+      return '';
+    }
+
+    if (maxDate > getEndOfDay(now)) {
+      return '';
+    }
+
+    return getShortTimeString(maxDate.getHours(), maxDate.getMinutes());
+  }, [maxDate, now]);
```
mdespuits, line 110: This effectively removes the need for `useMemo`. `now` will _always_ be a different value on each re-render so no memorization will ever happen. This may as well be a variable assignment or a normal function

src/components/InputTime/InputTime.tsx
```diff
@@ -25,12 +29,115 @@ Desired features:
...
+    if (minDate === undefined || minDate === null) {
+      return '';
+    }
+
+    if (minDate < getStartOfDay(now)) {
+      return '';
+    }
+
+    return getShortTimeString(minDate.getHours(), minDate.getMinutes());
+  }, [minDate, now]);
```
mdespuits, line 122: Same here as in `useMemo` for the max value

src/components/InputTime/InputTime.tsx
pixelbandito: Crud, good call.
I'm not sure how often to generate the date.
Upside of always re-calculating: If you're on the form at midnight, and your date suddenly becomes invalid, we catch it.

But it seems like a lot of recalculating and date parsing for very little benefit.

What do you think? Should I consider getting `now` just once on mount?

mdespuits: I think I would lean towards that.

However, it may make sense to floor to the nearest minute. That way if a few minutes pass during a restroom break or something the user can come back and it will be recalculated to the most up-to-date minute.

src/components/InputTime/InputTime.tsx
```diff
@@ -31,138 +18,31 @@ interface InputTimeProps extends Omit<InputProps, 'value' | 'max' | 'min'> {
...
   max,
   min,
-  onChange,
   value,
   ...passedProps
 }) => {
-  const mainRef = useRef<HTMLInputElement | null>(null);
-  const shadowTimeInputRef = useRef<HTMLInputElement | null>(null);
-
-  const syncValidity = () => {
```
pixelbandito, line 45: 90% of `InputTime` is now in `InputTime/AsString`

`InputTime` is now a wrapper than conditionally renders either `AsString` or `AsDate`

src/components/InputTime/AsDate.tsx
```diff
@@ -0,0 +1,162 @@
...
+  onChange,
+  value: valueDateString,
+  ...passedProps
+}) => {
+  const [invalidMax, setInvalidMax] = useState(false);
+  const [invalidMin, setInvalidMin] = useState(false);
+
+  useEffect(() => {
+    if (maxDateString && Number.isNaN(new Date(maxDateString).valueOf())) {
+      console.warn('`max` should be an ISO date string');
```
mdespuits, line 34: One quick thought: it might be useful to add a reference to `InputTime`. May help a bit in tracking down where the error is originating

PR #226, opened 2020-06-26
new svg icon for articleList

kcj002:
Adds new icon for list-alt-2 svg for use in widget edit form.

Not sure if warrants version bump. Plz send help.

### DIFF
--- a/src/components/Icon/icons/ArticleList.js
+++ b/src/components/Icon/icons/ArticleList.js
@@ -3,13 +3,15 @@ import PropTypes from 'prop-types';
 import iconColorsMap from './colors';
 
 const ArticleList = ({ fill, ...props }) => (
-  <svg width="24" height="24" viewBox="0 0 24 24" fill={fill} {...props}>
-    <rect x="2" y="5" width="6" height="5" />
-    <rect x="10" y="5" width="12" height="2" />
-    <rect x="10" y="8" width="8" height="2" />
-    <rect x="2" y="14" width="6" height="5" />
-    <rect x="10" y="14" width="12" height="2" />
-    <rect x="10" y="17" width="8" height="2" />
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width="24"
+    height="24"
+    viewBox="0 0 24 24"
+    fill={fill}
+    {...props}
+  >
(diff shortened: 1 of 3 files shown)


src/components/Icon/icons/ArticleList.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import iconColorsMap from './colors';
+
+const ArticleList = ({ fill, ...props }) => (
+  <svg width="24" height="24" viewBox="0 0 24 24" fill={fill} {...props}>
```
mdespuits, line 6: For all of our SVGs, we have run them through `svgo` to optimize every SVG in the library. One tool you can quickly use for this ishttps://jakearchibald.github.io/svgomg. You also need to include the `xmlns="http://www.w3.org/2000/svg"` attribute for consistency.

kcj002:
> I'm not entirely sure that this is a good fit for RoverUI since this is domain-specific. I would lean in one of two directions:
> 
> * Add this component to the FE app
> * Add this as an icon to tk-ui.
> 
> @pixelbandito any thoughts on this?

That gets complicated for use in the front end where we have constants for the icons that are all coming from rover. I could change the name of the icon so that it is less TK specific of a name.

mdespuits:
> I could change the name of the icon so that it is less TK specific of a name.

That seems like a reasonable thought.Maybe something like `mediaList`? I also am not completely opposed to the name `articleList` either. Just jumped out at me as a very Cision domain

src/components/Icon/icons/MediaList.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import iconColorsMap from './colors';
 
-const ArticleList = ({ fill, ...props }) => (
+const ListAlt2 = ({ fill, ...props }) => (
```
mdespuits, line 5: If you rename this to `MediaList`, the const should match the file name 😄

kcj002: ugh. Had the change but didn't save/commit it. it's not my day. 🤦‍♀️

pixelbandito:
Hella tight

PR #227, opened 2020-06-29
Use np package for releasing versions

mdespuits:
Our current workflow for releasing versions is not automated and currently fraught with possible issues (not running build before release, having local changes during release, etc).


mdespuits:
Closing in favor of relying on `npm version` and `yarn publish` flow that is less reliant on the need for being able to push to the main branch

mdespuits:
Closing in favor of relying on `npm version` and `yarn publish` flow that is less reliant on the need for being able to push to the main branch

PR #229, opened 2020-07-01
yarn new-version

mdespuits:
## Adds `yarn new-version`

[asciicast](https://asciinema.org/a/323LgEdFfv2jSuUjTOIknVFgx)

This calls `yarn version --no-git-tag-version`, which doesn't create a tag, and then auto-commits the changes with the appropriate commit message. This will allow the ability to tag the squashed merge commit in the `master` branch instead of a commit squashed in a PR.

Documents current workflow for releasing new versions of Rover for both Champions as non-Champions alike.


package.json
```diff
@@ -52,10 +52,13 @@
...
-    "prepublish": "in-publish && run-p test build || not-in-publish",
+    "prepublish": "in-publish && run-p -l test build || not-in-publish",
     "predeploy": "cd example && yarn install && yarn build",
-    "postpublish": "git tag v$npm_package_version && git push --tags && echo \"Successfully released version $npm_package_version!\"",
-    "release": "yarn publish --no-git-tag-version",
+    "postpublish": "git push --tags && echo \"Successfully released version $npm_package_version!\"",
+    "preversion": "run-p -l test build",
+    "version": "yarn update-changelog && git add CHANGELOG.md",
+    "new-version": "yarn version --no-git-tag-version",
+    "postnew-version": "git add . && git commit -m \"v$npm_package_version\" -n",
```
pixelbandito, line 61: The workflow is a little uneven in terms of what git commands are automatic (`git add . && git commit ...`) vs. manual (`git checkout master && git fetch -p && git reset ...`)

I'd probably err on the side of asking the user to do the `git` commands manually, at least at first; but it could go either way.

But I won't be a stickler about it.

CONTRIBUTING.md
```diff
@@ -59,12 +59,27 @@ Please tread carefully, and add any issues or suggestions on the [GitHub issue t
...
-
-If you don't have write access to the repo, one of the admins can do steps 3 and 4 for you.
+### Non-champions
+
+1. Checkout a release branch of some kind: (e.g. `release-2.0.4` or `new-v2.1.4`)
+2. **Following [semver](https://semver.org)** as your version guide, use the script `yarn new-version [value]` to bump the version. See [`yarn version`](https://classic.yarnpkg.com/en/docs/cli/version/) docs for more info
+   1. For `patch`: `yarn new-version --patch`
+   2. For `minor`: `yarn new-version --minor`
+   3. For `major`: `yarn new-version --major`
+   4. For other: `yarn new-version --new-version 3.4.11`
```
pixelbandito, line 69: Using this pattern, there won't be any kind of tag attached to the version, right?

I think I see why that's necessary, I just want to make sure I understand it.

package.json
mdespuits: Yeah, it's kinda trickier than I would have hoped. The main problem is that we do not allow anyone to push directly to master unless they are a champion (which makes sense).

The assumption of `yarn version` and other patterns is that this ability is present. It also assumes that the `tag` is on the final commit for the version. In our case, this can also be a squashed commit from a pull request, not the original commit itself. That's why there is a _slightly_ different flow for non-champions and champions and thus the usefulness of `new-version`.

CONTRIBUTING.md
mdespuits: Yes, with `new-version` it doesn't create the tag because otherwise, the tag will be on a commit that will be squashed from the PR and not exist in the `master` branch.

PR #232, opened 2020-07-06
Add timepicker dropdown and clock icon

pixelbandito:


src/components/Button/Button.tsx
```diff
@@ -11,21 +11,27 @@ import { TButtonLevel, TButtonSize, TButtonState } from './types';
...
-  tag?: React.ComponentType<{ [key: string]: unknown }> | null;
+  tag?: React.FC<TagType> | React.ComponentType<TagType> | null;
 
   // States
   hover?: boolean;
   active?: boolean;
   disabled?: boolean;
   checked?: boolean;
   focus?: boolean;
+  [key: string]: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
pixelbandito, line 34: TS was being a jerk about letting you pass arbitrary props down when you send a custom tag (like `"div"`, `"a"`, or `MyComponent`).

src/components/InputTime/InputTime.module.css
```diff
@@ -0,0 +1,29 @@
+.InputTime {
+  position: relative;
+  display: inline-flex;
+  flex-flow: row nowrap;
+  padding: 0;
+}
+
+.Input.Input {
```
pixelbandito, line 8: This suppresses the nested `Input`'s non-layout-related styles so that the `InputTime` styles are the only border / background / etc.

src/components/Input/Input.module.css
```diff
@@ -35,7 +35,8 @@
   color: var(--rvr-gray-80);
 }
 
-.Input:focus {
+.Input:focus,
+.Input:focus-within {
```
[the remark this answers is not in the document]
sebastianvera: This is not supported by IE11, is that ok?

src/components/InputTime/AsString.tsx
```diff
@@ -50,6 +56,10 @@ const AsString: React.FC<AsStringProps> = ({
   >;
 
   useImperativeHandle(forwardedRef, () => localRef.current, []);
+  const [validity, setValidity] = useState(true);
+
+  const showDropdown: 'click' | undefined =
+    (customShowDropdown === 'click' && 'click') || undefined;
```
sebastianvera, line 62: nit: isn't this the same as `customShowDropdown === 'click' ? 'click' : undefined`? I think it might be confusing to have the `&& 'click'` right after checking that they have the same value

src/components/InputTime/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,152 @@
...
+    } while (attempts <= maxAttempts && current < end);
+
+    return options;
+  }, [customStep, max, min, value, onSelectMenuItem, showDropdown]);
+
+  return (
+    <EasyDropdown
+      className={classNames(className, styles.Dropdown)}
+      disabled={disabled}
+      isOpen={undefined}
```
sebastianvera, line 124: if this props is not speficied, is it the same as passing `undefined`?

src/components/InputTime/InputTime.test.tsx
```diff
@@ -33,6 +33,86 @@ describe('InputTime', () => {
     expect(baseInputTime.value).toBe('');
   });
 
+  describe.only('dropdown time picker', () => {
```
sebastianvera, line 36: I believe you forgot to remove the `.only` calls here

mdespuits: Maybe add the `eslint` rule to avoid committing these?

src/components/Button/Button.tsx
mdespuits: Well, the idea was to limit `<Button>` to be an actual button or anchor tag. `button` and `a` tags, as well as React components, make sense.

What was the problem that required this change? This effectively allows "anything" to be passed and therefore anything that is not a proper HTML attribute will get rendered along with the element. Does that make sense?

src/components/InputTime/AsString.tsx
mdespuits: I would agree with this. Seems like a tricky trick that would possibly confuse future developers (myself included!)

## 🧑‍⚖️

src/components/Input/Input.module.css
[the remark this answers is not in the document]
pixelbandito: Switched to tracking local focus state. 😢

src/components/InputTime/AsString.tsx
pixelbandito: Word.

src/components/Button/Button.tsx
pixelbandito: iirc, we were spreading arbitrary tags onto the Button on purpose.

Not only can you do `aria-label="foo"` (and a number of other arbitrary html attributes), you also can pass a component as a tag, like `tag={NavLink}`, in which case any valid prop for the component is a valid prop for that instance of `Button`.

src/components/InputTime/Dropdown/Dropdown.tsx
```diff
@@ -121,7 +121,9 @@ const Dropdown: React.FC<DropdownProps> = ({
     <EasyDropdown
       className={classNames(className, styles.Dropdown)}
       disabled={disabled}
-      isOpen={undefined}
+      isOpen={
+        undefined /* Suppresses typescript lint warning about missing prop */
+      }
```
mdespuits, line 126: Why does `isOpen` need to be `undefined` instead of `false`?

pixelbandito: Added a comment to clarify this.
It's because of the TS component importing a JSX component. In this case, omitting `isOpen` was inaccurately throwing a missing prop error. But `EasyDropdown` is specifically looking for isOpen to be undefined - if it's undefined, the component uses an uncontrolled open state.

I could have also suppressed the TS warning by editing `EasyDropdown` to have 
```
const EasyDropdown = ({
  ...
  isOpen: controlledIsOpen = undefined,
  ...
}) => {
```

But that seemed just as weird. I figure porting EasyDropdown to TS will happen later, and then this will be a non-issue.

src/components/InputTime/Dropdown/Dropdown.tsx
pixelbandito: Added a comment to hopefully clarify.

PR #236, opened 2020-07-13
`InputTime` takes step prop for dropdown options and validation

pixelbandito:


src/components/InputTime/AsString.tsx
```diff
@@ -65,6 +66,34 @@ const AsString: React.FC<AsStringProps> = ({
   // value is the true, controlled value, patched for undefined / empty
   const value = useMemo(() => valueOrUndefined || '', [valueOrUndefined]);
 
+  const step: number | undefined = useMemo(() => {
+    let safeStep: number | undefined;
+    safeStep = typeof customStep === 'number' ? customStep : safeStep;
+
+    safeStep =
```
pixelbandito, line 73: Native HTML input `step` attributes are numbers as strings, so React's built-in Input types allow both.
I'm playing nice by parsing both.

src/components/InputTime/AsString.tsx
```diff
@@ -65,6 +66,34 @@ const AsString: React.FC<AsStringProps> = ({
   // value is the true, controlled value, patched for undefined / empty
   const value = useMemo(() => valueOrUndefined || '', [valueOrUndefined]);
 
+  const step: number | undefined = useMemo(() => {
+    let safeStep: number | undefined;
+    safeStep = typeof customStep === 'number' ? customStep : safeStep;
```
sebastianvera, line 71: this looks a little bit confusing, what about using a switch statement instead?

src/components/InputTime/AsString.tsx
```diff
@@ -65,6 +66,34 @@ const AsString: React.FC<AsStringProps> = ({
...
+
+    safeStep =
+      typeof customStep === 'string' ? parseInt(customStep, 10) : safeStep;
+
+    if (
+      !safeStep ||
+      Number.isNaN(safeStep) ||
+      safeStep < 60 * 10 ||
```
pixelbandito, line 79: Hmm, maybe I should change this bit.
Steps greater than a day are obviously garbage, but...

For the dropdown, we won't use steps less than 10 minutes because menu lengths.
But for actual input validation, I guess 2 minute, 5 minute steps would be valid.

I'll fix (by making a separate dropdown menu step from the `safeStep`

src/components/InputTime/AsString.tsx
pixelbandito: This section was pretty pointless. I'm going to just `parseInt(customStep, 10)` for all cases.

`parseInt(customStep, 10)`  works on numbers just fine.
`parseInt(undefined, 10)` produces `NaN`, which is explicitly guarded against in the next block.

pixelbandito:
I want to add some more tests around the step functionality, but I'd like to defer that to a later PR.

src/components/InputTime/AsString.tsx
pixelbandito: Would you have another look when you get a chance, @sebastianvera?

package.json
```diff
@@ -44,8 +44,8 @@
     "pre-commit:lint": "yarn lint-staged",
     "pre-commit:types": "yarn tsc",
     "pre-commit:test": "yarn test",
-    "test": "cross-env CI=1 react-scripts test --env=jsdom",
-    "test:watch": "react-scripts test --env=jsdom",
+    "test": "cross-env CI=1 react-scripts test --env=jsdom-fourteen",
+    "test:watch": "react-scripts test --env=jsdom-fourteen",
```
mdespuits, line 48: What is this change?

pixelbandito: Oh, when I was trying to test with `waitFor`, I discovered the react-scripts default version of jsdom doesn't support a required mutation observer.

However, I don't have `waitFor` in the tests anymore, so I'm reverting this.

PR #237, opened 2020-07-13
`InputTime` takes `formatTime` for custom i18n time string formatting

pixelbandito:


pixelbandito:
I did a quick-and-dirty test integration with `yalc` on the `trendkite-front-end` app.
Passing `(date) => iintl.formatTime(date, { hour: 'numeric', minute: 'numeric', timeZoneName: 'short' })` as the `formatTime` prop works as expected.

PR #238, opened 2020-07-16
Fix Pill action spacing

dkordik:
See commits for granular change specifics.

## Before
image

## After
image


dkordik:
aw oops, conflict, looks like I didn't have the latest copy of master. will have to jump back into fixing that next week

PR #243, opened 2020-07-24
ts conversion in one fell swoop--hopefully

droddy:


src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,48 @@
...
+  useState,
+  ReactElement,
+  FC,
+  SyntheticEvent,
+  Children,
+  cloneElement,
+} from 'react';
+
+interface AccordionProps extends React.HTMLAttributes<HTMLDivElement> {
+  children?: any | null;
```
droddy, line 11: derp! just read the component readme

> This components takes an arbitrary amount of `ExpansionPanel` components as children

I'll go ahead and set that as the children type

src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,48 @@
...
+  defaultExpandedPanel?: number | null;
+}
+
+const Accordion: FC<AccordionProps> = ({
+  children,
+  defaultExpandedPanel,
+  ...rest
```
pixelbandito, line 18: Minor, naming thing: The word "rest" is like writing a function that takes "arg" - no semantic benefit. That's why I prefer `passedProps` (unless you're doing something other than passing the extra props down to another node).

src/components/Accordion/Accordion.tsx
pixelbandito, line 11: There's a ReactNode type, I think. I might have the name wrong.

droddy: Yeah, I was trying that but ran into other TS issues I could not resolve. After I convert children arg to the ExpansionPanel I'll try again. Thanks.

droddy: ReactNode was right...then had to cast to React.Element array to successfully iterate over it.

droddy: stuck with more generic ReactNode

PR #244, opened 2020-07-27
refactor: move sidetray to typescript

### LINKED ISSUE #197: Convert SideTray to TypeScript

NicolasBiehler:
Fixes #197


src/components/SideTray/SideTray.tsx
```diff
@@ -7,7 +7,65 @@ import { parseCssSize } from '../../shared/css-utils';
...
+
+interface ChildComponentsProps {
+  children: ReactNode;
+  className?: string;
+}
+
+const Header: React.FC<ChildComponentsProps> = ({ className, ...props }) => (
+  <div {...props} className={`${styles.header} ${className}`} />
+);
+Header.propTypes = {
```
pixelbandito, line 38: Can you drop the Header/Footer/Body propTypes now?

src/components/SideTray/SideTray.tsx
```diff
@@ -7,7 +7,65 @@ import { parseCssSize } from '../../shared/css-utils';
...
+  children: PropTypes.node.isRequired,
+  className: PropTypes.string,
+};
+Header.defaultProps = { className: '' };
+
+const Footer: React.FC<ChildComponentsProps> = ({ className, ...props }) => (
```
pixelbandito, line 44: When we switch to TS, we generally remove the defaultProps and declare default function parameters instead.
Badge.tsx is a pretty good reference, iirc.

src/components/SideTray/SideTray.tsx
```diff
@@ -7,7 +7,65 @@ import { parseCssSize } from '../../shared/css-utils';
...
+interface SideTrayStyles {
+  bottom: number | string;
+  height: string;
+  left: number | string;
+  right: number | string;
+  top: number | string;
+  width: string;
+}
+
+interface ChildComponentsProps {
```
pixelbandito, line 30: For clarity, maybe rename this `SideTrayChildProps` ?

src/components/SideTray/SideTray.tsx
```diff
@@ -7,7 +7,65 @@ import { parseCssSize } from '../../shared/css-utils';
...
+  className?: string;
+  direction?: string;
+  height?: string | number;
+  onClose: (...args) => void;
+  style?: object;
+  visible?: boolean;
+  width?: string | number;
+}
+
+interface SideTrayStyles {
```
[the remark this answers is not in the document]
NicolasBiehler: Yes I was not happy with that part, we talked about it with Seba but I knew no better. Thanks for that, I will change it now.

src/components/SideTray/SideTray.tsx
```diff
@@ -1,20 +1,54 @@
...
 import { parseCssSize } from '../../shared/css-utils';
 
 // This component's dependencies
 import styles from './SideTray.module.css';
 
-const SideTray = ({
+interface SideTrayProps {
+  children: ReactNode;
+  className?: string;
+  direction?: string;
```
mdespuits, line 14: A suggestion to set limit the number of options you can pass to `direction`: like

```ts
type SideTrayDir = 't' | 'b' | 'r' | 'l';

interface SideTrayProps {
  // ...
  direction?: SideTrayDir;
  // ...
}
```

That way it's enforceable at a type level and not just run-time

NicolasBiehler: True, let me add this real quick and this will be all good.

PR #250, opened 2020-08-03
refactor: Convert EasyPill component to typescript

bkonuwa:
Convert EasyPill component from js to ts
Added CSS Module file to style the drop down delete icon in Easy Pill


src/components/EasyPill/EasyPill.module.css
```diff
@@ -0,0 +1,4 @@
+.dropdownDeleteIcon {
+  align-self: center;
+  margin-right: var(--rvr-space-md);
+}
```
pixelbandito, line 4: Minor, but you should leave a blank newline at the end of the file.
It's a common thing with files under git version control, your editor likely has a setting for adding newlines at the end of files.

src/index.ts
```diff
@@ -14,7 +14,7 @@ export { default as Collapse } from './components/Collapse';
 export { default as DeletablePill } from './components/DeletablePill';
 export { default as Dropdown } from './components/Dropdown';
 export { default as EasyDropdown } from './components/EasyDropdown';
-export { default as EasyPill } from './components/EasyPill';
+export { default as EasyPill } from './components/EasyPill/EasyPill';
```
pixelbandito, line 17: You shouldn't need the extra `/EasyPill`
The `EasyPill/index` file re-exports the default from `./components/EasyPill`

src/components/EasyPill/EasyPill.tsx
```diff
@@ -1,13 +1,36 @@
...
-  if (!actions.length) {
+type EasyPillActions = {
+  children?: React.ReactNode;
+  label: string;
+  onClick: () => void;
+  group?: string;
+};
+
+interface EasyPillProps {
+  actions: EasyPillActions[];
```
pixelbandito, line 17: `actions` should be optional in the main `EasyPill` interface.

src/components/EasyPill/EasyPill.tsx
```diff
@@ -1,13 +1,36 @@
...
+  checked?: boolean;
+  children: React.ReactNode;
+  onClick?: () => void;
+  onDelete?: (...args) => void;
+}
+
+interface EasyPillDropdownProps {
+  actions: EasyPillActions[];
```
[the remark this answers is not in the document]
bkonuwa: You are correct, it should be optional. I will fix that. Thanks for catching it.

src/components/EasyPill/EasyPill.tsx
```diff
@@ -1,13 +1,36 @@
...
+  label: string;
+  onClick: () => void;
+  group?: string;
+};
+
+interface EasyPillProps {
+  actions: EasyPillActions[];
+  checked?: boolean;
+  children: React.ReactNode;
+  onClick?: () => void;
```
pixelbandito, line 20: This component doesn't a top-level `onClick`, right? I think this prop can go away.

src/components/EasyPill/EasyPill.tsx
```diff
@@ -1,13 +1,36 @@
...
 import Icon from '../Icon';
 import Pill from '../Pill';
+import styles from './EasyPill.module.css';
 
-const EasyPillDropdown = ({ actions, onDelete }) => {
-  if (!actions.length) {
+type EasyPillActions = {
+  children?: React.ReactNode;
+  label: string;
+  onClick: () => void;
```
pixelbandito, line 12: This click event should be super permissive, since it's not used at all in this component, just passed down to a child.
I think the built-in `Function` type would work.
Then, any type checking can happen in the `EasyDropdown` component that receives the prop.

src/components/EasyPill/EasyPill.test.tsx
```diff
@@ -2,17 +2,29 @@ import React from 'react';
...
+  const defaultActions = [
+    {
+      label: 'myAction',
+      onClick: () => {},
+    },
+  ];
   it('renders', () => {
-    shallow(<EasyPill onDelete={() => {}}>EasyPill</EasyPill>);
+    shallow(
+      <EasyPill actions={defaultActions} onDelete={() => {}}>
```
pixelbandito, line 16: I saw Dan's comment - he's right, `actions` wasn't required before. We should leave it that way, and then this won't be necessary.

src/components/EasyPill/EasyPill.tsx
bkonuwa: @pixelbandito I added that `onClick` prop because in the original implementation of the unit test, we are passing onClick to the EasyPill component and I was seeing this TS error `Property 'onClick' does not exist on type 'IntrinsicAttributes & EasyPillProps & { children?: ReactNode; }'`
So should I remove the `onClick` prop from the unit test as well?

src/components/EasyPill/EasyPill.tsx
bkonuwa: Just for clarity, do you mean change `onClick: () => void;` to `onClick: Function;`?

src/components/EasyPill/EasyPill.tsx
pixelbandito: I think that would be good. No need to test a prop that doesn't isn't used by the component.

While looking at this, I realized that the `...passedProps` for EasyPill` are passed to a `Pill` component.
`Pill` is already using TS, so you should change the `EasyPillProps` interface to extend `PillProps` from that component.

Something like
```
import { PillProps } from '../Pill/Pill';

// ...

interface EasyPillProps extends PillProps {
  // ...
}
```

That will give you the `checked` prop for free, and it will allow consumers to pass any additional props down that are valid for `Pill`.

PR #252, opened 2020-08-03
feat: InputTime dropdown menu has default max height of 10.5 rows

### LINKED ISSUE #249: `InputTime` dropdown scrolling
Relates to #203 

- dropdown has a max-height and internal scrolling

pixelbandito:
fixes #249

Feature:
- Re-using existing functionality to pass `menuProps` to `EasyDropdown`.
 - `InputTime` takes `dropdownProps` which are passed to `InputTime/Dropdown`.
 - `InputTime/Dropdown` passes any `menuProps`, `toggleProps`, or top-level `passedProps` down to its children in the appropriate slots.
- By default, `InputTime/Dropdown` adds a CSS class to the internal `Dropdown.Menu` which limits the height to 5.5 rows.
- `Dropdown.Menu` has a default style of `overflow-y: auto` for scrolling.

Docs:
- Added some stories to document the new ability.

Fix:
- Fixed `InputTime`'s clock icon to use correct color when disabled.
- Fixed `InputTime`'s `fauxDisabled` prop to pass it down to the correct child, so that class has the desired visual effect.

### DIFF
--- a/src/components/InputTime/Dropdown/Dropdown.module.css
+++ b/src/components/InputTime/Dropdown/Dropdown.module.css
@@ -6,9 +6,13 @@
   margin-right: 0;
 }
 
-.dropdownToggle {
+.toggle {
   display: block;
   padding: 3px;
+}
+
+.menu {
+  max-height: calc(var(--rvr-space-lg) * 2 * 5.5); /* --rvr-space-lg * 2 is the height of one row. */
 }
 
 .menuItem {
--- a/src/components/InputTime/Dropdown/Dropdown.module.css
+++ b/src/components/InputTime/Dropdown/Dropdown.module.css
@@ -12,7 +12,9 @@
 }
 
 .menu {
-  max-height: calc(var(--rvr-space-lg) * 2 * 5.5); /* --rvr-space-lg * 2 is the height of one row. */
+  --item-height: calc(var(--rvr-space-lg) * 2);
+  --visible-items-before-scrolling: 5.5;
+  max-height: calc(var(--item-height) * var(--visible-items-before-scrolling));
 }
 
 .menuItem {
--- a/src/components/InputTime/Dropdown/Dropdown.tsx
+++ b/src/components/InputTime/Dropdown/Dropdown.tsx
@@ -39,12 +39,15 @@ const Dropdown: React.FC<DropdownProps> = ({
   disabled,
   formatTime,
   max,
+  menuProps,
   min,
   onSelectMenuItem,
+  passedProps,
   showDropdown,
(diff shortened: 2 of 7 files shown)


src/components/InputTime/Dropdown/Dropdown.module.css
```diff
@@ -6,11 +6,15 @@
...
 }
 
-.dropdownToggle {
+.toggle {
   display: block;
   padding: 3px;
 }
 
+.menu {
+  max-height: calc(var(--rvr-space-lg) * 2 * 5.5); /* --rvr-space-lg * 2 is the height of one row. */
```
dkordik, line 15: might we bring more clarity to this line by putting `2` and `5.5` into some well-named variables? (might not need this comment at that point)

src/components/InputTime/Dropdown/Dropdown.tsx
```diff
@@ -24,12 +24,14 @@ interface DropdownProps {
...
+  menuProps?: Record<string, any>; // eslint-disable-line @typescript-eslint/no-explicit-any
   min?: string;
   onSelectMenuItem: Function;
   showDropdown?: 'click' | 'focus';
   step?: number;
   stepFrom?: string;
   toggleAriaLabel?: string;
+  toggleProps?: Record<string, any>; // eslint-disable-line @typescript-eslint/no-explicit-any
   value?: string;
   [key: string]: any; // eslint-disable-line @typescript-eslint/no-explicit-any
```
dkordik, line 36: looks like this isn't the first time we've overridden this rule. should we just allow explicit any's project wide?

src/components/InputTime/Dropdown/Dropdown.module.css
pixelbandito: Just local variables for clarity? 💯 
I'll be more nervous if you mean global variables. This isn't a universal menu item height, it's reverse-engineered from this particular dropdown usage.

dkordik: just local, ya

src/components/InputTime/Dropdown/Dropdown.tsx
pixelbandito: I think it's still useful.
If the consumed components had been TS already, I would have used those instead of `Record<string, any>`.
As it is, I'd rather wait for the conversion than re-type them here.

src/components/InputTime/Dropdown/Dropdown.module.css
pixelbandito: Done

Issue #248, opened 2020-07-27
`InputTime` picker should have a timezone offset prop


pixelbandito:
The `InputTime` picker currently uses a locale-based start of day to calculate:

- the options in the timepicker dropdown
- validation against the `step` prop. (Valid times are start of day + increments of `step` in seconds, if no `min` prop is provided.

For i18n concerns, those use cases should respect a configured, numeric timezone offset property.

There might be something smart we can do with the default time formatter as well - it controls the format of valid times in the fuzzy input and the dropdown options. But if tz offset isn't applicable there, there's an existing workaround for consumers by using a custom `formatTime` function.

InputTime takes user input "fuzzy" times and parses them to 24-hour time strings or ISO date strings.

 

It does some validation and also renders a dropdown with "smart" options.

The validation and dropdown options currently rely on user local time, but they should accept a timezone offset.

 
**Currently:**
If ...
- the time formatter is parsing to "India Standard Time" (5:30 hours ahead of UTC)
- the browser is somewhere in UTC timezone
- there's not a "min" time provided

... then ...
- the dropdown will render options starting from a weird half-hour
- the options provided will start at "5:30 PM" (12:00AM UTC, translated into IST).

 **Desired:**
- the dropdown options start at "12:00 AM"
- selecting "12:00AM" will map to 12:00 IST
- if `step` is set to an hour, 12:00 IST will be a valid time, and 12:30 IST will not.

PR #255, opened 2020-08-10
test: Add `InputTime/utils` tests and prepare for `timeZoneOffset` prop

pixelbandito:
Tests cover behavior with and without `timeZoneOffset`

`timeZoneOffset` isn't used yet, but this prep work is intended to make its addition more seamless.

relates to #248

### DIFF
--- a/src/components/InputTime/AsString.tsx
+++ b/src/components/InputTime/AsString.tsx
@@ -91,7 +91,7 @@ const AsString: React.FC<AsStringProps> = ({
   }, [customStep]);
 
   const dropdownStep: number | undefined = useMemo(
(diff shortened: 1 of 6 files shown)


src/components/InputTime/AsString.tsx
```diff
@@ -91,7 +91,7 @@ const AsString: React.FC<AsStringProps> = ({
   }, [customStep]);
 
   const dropdownStep: number | undefined = useMemo(
-    () => (step && step >= 60 * 10 ? step : undefined),
+    () => (step && step >= SECONDS_PER_MINUTE * 10 ? step : undefined),
```
sebastianvera, line 94: clean code? 😛

PR #257, opened 2020-08-17
Input time with tz offset

pixelbandito:
The `InputTime` component lets you pick a time of day.
Depending on the props passed in, it uses a min/max/value of either simple time of day strings ("10:00") or ISO date strings.

It validates selections against the `min` and `max`, and it also creates a dropdown menu with pre-populated options based on them.

The new functionality lets you pass in a target `timeZoneOffset`, e.g. the user is selecting times for IST (India Standard Time), which is +5:30 from UCT.
* The functions that detect start of day, end of day to build dropdown options do the right thing.
* We correctly determine if the `min` / `max` dates (in date mode) encompasses the selected date in India, so the validation can detect whether all or only some times are valid for a passed-in ISO date string.
* The configured `step` prop correctly validates using India's `min` or start of day. So 10:30 UTC is a valid option for India with no `min` and a `step` of 1 hour.

### DIFF
--- a/src/components/InputTime/AsDate.tsx
+++ b/src/components/InputTime/AsDate.tsx
@@ -148,6 +148,17 @@ export const AsDate: React.FC<AsStringProps> = ({
       nextDate.setSeconds(0);
       nextDate.setMilliseconds(0);
 
+      /*
+        When the user is in a different time zone than specified by
+        `timeZoneOffset`, the valid options for single date will appear to be on
+        different calendar dates.
+
+        E.g. For a user in UTC, India's "04:30" and "05:30" options will be on
+        different dates.
+
+        Since the options never span more than 24 hours, this cheesy logic
+        works to force the selection into the correct date.
+      */
       if (nextDate > maxDate) {
         nextDate.setDate(nextDate.getDate() - 1);
       }
--- a/src/components/InputTime/AsDate.tsx
+++ b/src/components/InputTime/AsDate.tsx
@@ -9,6 +9,7 @@ import React, {
 } from 'react';
 
 import {
+  getDateTimeFromShortTimeString,
   getEndOfDay,
   getShortTimeString,
   getStartOfDay,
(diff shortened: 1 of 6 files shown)


wendigolabs:
Just a general question about timezone handling in Rover: Is it possible to get `moment.tz` to do any of the heavy lifting?

pixelbandito:
@wendigolabs 

> Just a general question about timezone handling in Rover: Is it possible to get `moment.tz` to do any of the heavy lifting?

I'm sure. It probably would have made things easier, but:
* Moment and moment.tz are big libraries.
* This might be way off base, but I think moment.tz requires the time zone name ("America/Chicago") and would be harder to use that a minute-based `timeZoneOffset`. I didn't want to force consumers to use something heavy like moment if they didn't want to.

That said, if I find any more tricky edge cases with this thing, I'll re-evaluate. This was way harder to do than I expected.

pixelbandito:
Found another bug. A small one, I think, but I'll be pushing a change to this branch soon (tomorrow) to address it.

dkordik:
maybe it's early enough in this component's life to suggest this- would we want the time picker to only deal in dumb "minute"/"hour"/"second" inputs and outputs, instead of actual Dates? that would leave it on the implementor to convert to/from Date objects if they so desired (and do any timezone calculation necessary to their use case) and leave this component pretty much just presentational

pixelbandito:
@dkordik 

> maybe it's early enough in this component's life to suggest this- would we want the time picker to only deal in dumb "minute"/"hour"/"second" inputs and outputs, instead of actual Dates? that would leave it on the implementor to convert to/from Date objects if they so desired (and do any timezone calculation necessary to their use case) and leave this component pretty much just presentational

That's more or less what the internal "AsString" component does (with the caveat that we deemed seconds irrelevant to our foreseeable use cases.

I think I'll expose the simpler `AsString` as an export, but I'd like to keep offering the more robust version as well.

Hope I'm not falling for the sunk cost fallacy here. The fact that it's been this hard for me to get right makes me think it will be valuable to the consumers.

mikemfleming (approved):

pixelbandito:
This one looked like it worked, but there were still timezone issues. Closing the PR until further notice.

PR #258, opened 2020-08-17
fix: min max validation without native input[type=time] support

pixelbandito:
Targeting a WIP branch for now, but this single commit should carry over easily after that's merged.

Safari doesn't support input with type="time" natively, alone among major browsers. That means the native validation wasn't working.

I wrote a pretty cheesy shim, at the basic `AsString` component, so `AsDate` should be fixed for free, and this doesn't require any time zone consideration.


pixelbandito:
I'm merging a version of this fix directly into master, before extended timezone offset support is available.

PR #260, opened 2020-08-20
Input time with tz offset pt 2

pixelbandito:
- fixes an issue with min/max validation in the `AsDate` component.
- refactor: changes the interface for a few of the `InputTime/utils`, so there's a slight refactor there.
- Adds named export for `AsDate`: `InputTimeAsDate`
- Adds named export for `AsString`: `InputTimeAsString`
- docs: Most of the huge diff comes from updating the `InputTime` story to add sections for the `AsDate`/`AsString` components.

### DIFF
--- a/src/components/InputTime/utils.test.tsx
+++ b/src/components/InputTime/utils.test.tsx
@@ -360,19 +360,20 @@ describe('getDateTimeFromShortTimeString', () => {
   it('turns hour and minute into a local date', () => {
     let date: Date;
 
-    date = getDateTimeFromShortTimeString('12:30');
+    date = getDateTimeFromShortTimeString({ time: '12:30' });
     expect(date.toLocaleTimeString('en-US')).toEqual('12:30:00 PM');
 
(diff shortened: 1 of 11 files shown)


src/components/InputTime/utils.test.tsx
```diff
@@ -381,7 +382,8 @@ describe('getDateTimeFromShortTimeString', () => {
         timeZoneName: 'America/Chicago',
       });
 
-      const dateIST = getDateTimeFromShortTimeString('12:30', {
+      const dateIST = getDateTimeFromShortTimeString({
+        time: '12:30',
         timeZoneOffset: +330,
       });
```
dkordik, line 389: some good edge cases to add tests for:
- offset brings you into the previous day
- offset brings you into the next day
- offset brings you into the next day during daylight savings time? :shudder:

pixelbandito: Curses. That is a very good idea.

pixelbandito:
Hard nope until some underlying kinks are worked out.

Issue #267, opened 2020-09-03
Using custom CSS props in a file not included by postcss


pixelbandito:
InputTime/Dropdown/Dropdown.module.css defines some CSS vars, but isn't included in the postcss file list to read those vars.
This will work in modern browsers, but break in ones that don't support css vars.

I'd suggest we add the relevant vars to src/shared/sizing.css (if we can make them generic) or remove them altogether in favor of hard coding values. Adding arbitrary module.css files to postcss would be the alternative, but is a worse idea imo.

Issue #268, opened 2020-09-03
EasyTab doesn't allow readonly tab prop


pixelbandito:
Problem:
So far, only 1 consumer has had this issue:
Consumer added `EasyTabMenu` with a `tabs` prop that was marked `readonly`, but otherwise compatible with our types.

Proposed fix:
Make the `tabs` prop readonly.

---

This is maybe a foreshadowing of a broader change.
Most of the typed props we take from consumers aren't marked readonly currently. Maybe all of them should be.

PR #269, opened 2020-09-03
TypeScript and CSS custom prop fixes

pixelbandito:
Fixes issues #267 and #268, plus a couple of undocumented bugs with the move to TS.

- Some CSS vars moved to shared sizing.css
- EasyPill handles undefined `children` prop
- EasyTabMenu allows readonly tabs prop, and handles undefined items in the tabs array
- Export SimpleTabMenu as an alias of EasyTabMenu for backwards compatibility 

This will need a new patch version before we can upgrade Cision FE code to use RoverUI again.

### DIFF
--- a/src/components/TabMenu/TabMenu.tsx
+++ b/src/components/TabMenu/TabMenu.tsx
@@ -74,6 +74,6 @@ export const EasyTabMenu: React.FC<EasyTabMenuProps> = ({
   );
 };
 
-export const SimpleTabMenu = TabMenu;
+export const SimpleTabMenu = EasyTabMenu;
 
 export default TabMenu;
--- a/src/components/TabMenu/TabMenu.tsx
+++ b/src/components/TabMenu/TabMenu.tsx
@@ -32,7 +32,7 @@ export type EasyTabType = {
   onClick: (e: React.SyntheticEvent) => void;
 };
 export interface EasyTabMenuProps extends TabMenuProps {
-  tabs?: EasyTabType[];
+  tabs?: Readonly<EasyTabType[]>;
   size?: 'xs' | 'sm' | 'md' | 'lg';
   activeTab?: string;
 }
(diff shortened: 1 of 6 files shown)


nvmusoke (approved):

src/components/TabMenu/TabMenu.tsx
```diff
@@ -43,9 +44,17 @@ export const EasyTabMenu: React.FC<EasyTabMenuProps> = ({
   size = 'md',
   ...props
 }) => {
+  const safeTabs: EasyTabType[] = [];
+
+  tabs.forEach((tab) => {
+    if (tab) {
+      safeTabs.push(tab);
+    }
+  });
```
dkordik, line 53: might be simpler logic to filter by truthyness
```
tabs.filter(tab => !!tab)
.map..
```

dkordik (approved):

src/components/TabMenu/TabMenu.tsx
pixelbandito: I tried, but TS didn't like turning a list of `EasyTabType | undefined` tab into a list of `EasyTabType`, even with the filtering.

Maybe I was missing something obvious?

image

pixelbandito: BTW, I tried doing it inline with the `.map` first, then broke it out into its own const to verify what was going on.

pixelbandito: Ah, I'm not the first! Found a maybe better way.
https://stackoverflow.com/a/46700791

dkordik: oh man, sorry for the TS rabbit hole

src/components/TabMenu/TabMenu.tsx
```diff
@@ -42,10 +43,13 @@ export const EasyTabMenu: React.FC<EasyTabMenuProps> = ({
...
   ...props
-}) => {
-  return (
-    <TabMenu {...props}>
-      {tabs.map((tab) => {
+}) => (
+  <TabMenu {...props}>
+    {tabs
+      .filter(
+        (tab: EasyTabType | undefined): tab is EasyTabType => tab !== undefined
```
sebastianvera, line 50: neat!

PR #272, opened 2020-10-11
Removed third-party theming systems from the project.

wendigolabs:


mdespuits (approved):
🍔🌹

wendigolabs:
Ahhh, I just noticed that `Grid.test.js` is still referencing theming, so I need to fix that.  Kinda seems like that test should've failed?  But then again there's a test farther down that seems to bless passing an unparsed `gutter` value through:
```
  describe('with garbage string in `gutter` prop', () => {
    it('renders', () => {
      mount(<Grid gutter="garbage-string" />);
    });
  });
```
?

pixelbandito (approved):

PR #280, opened 2020-12-07
feat: Adding Typography component

bkonuwa:


src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,60 @@
...
+        styles[color],
+        styles[weight],
+        styles[align],
+        styles[size],
+        passedClassName
+      ),
+    [align, color, passedClassName, weight, size]
+  );
+
+  return <Tag {...props} className={className} />;
```
mikemfleming, line 54: This is interesting! Does this work because html element names are actually calls to `createElement` in jsx? 

What happen if someone passes in an invalid element name or one that isn't supported?

src/components/InputTime/utils.test.tsx
```diff
@@ -118,7 +118,7 @@ describe('getEndOfDay', () => {
     expect(endOfDay.toISOString()).toEqual(expectedDateString.toISOString());
   });
 
-  describe('with optional `timeZoneOffset`', () => {
+  describe.skip('with optional `timeZoneOffset`', () => {
```
mikemfleming, line 121: Why are these tests being skipped?

mikemfleming (commented):
This looks cool! I left some questions for you but it looks good to me.

src/components/Typography/Typography.tsx
pixelbandito: I _think_ it works that way, but I didn't read the source.

I first stole if from [reactstrap](https://github.com/reactstrap/reactstrap/blob/master/types/lib/Button.d.ts) (React Bootstrap). I love it for things like buttons that can be links, or typography that can be just about anything. imo, nice to read, super flexible.

It can be typed with [React.ElementType](https://flow.org/en/docs/react/types/#toc-react-elementtype)

src/components/Typography/Typography.tsx
```diff
@@ -0,0 +1,60 @@
+import React, { useMemo } from 'react';
+import classNames from 'classnames';
+
+import styles from './Typography.module.css';
+
+type TagType =
```
pixelbandito, line 6: Would it work to use [React.ElementType]( https://flow.org/en/docs/react/types/#toc-react-elementtype) here?

src/components/InputTime/utils.test.tsx
pixelbandito: This is my bad.

TLDR: I've been working on timeZoneOffset for a while - it's now on the backburner and not fully functional.

I have a long-running WIP to add a timezone offset to the `InputTime` picker.
Ideally, it would let consumers do something like 
```
<InputTime
  min="2020-12-07T22:33:05.925Z" 
  timeZoneOffset={-330}
/>
```
So they'd get a dropdown list starting with the time of day for the min date, and ending at midnight in the chosen location. I wrote it without using a library, and it's been a nightmare.

I keep finding edge cases, for instance, now that daylight savings is over, the tests fail.

I told Boima to ignore it for now, I need to circle back and fix it or delete the code.

src/components/Typography/README.md
```diff
@@ -0,0 +1,16 @@
+# \<Typography /\>
+
+### Typography is a helper component used to present contents clearly and uniformly
+
+By default, it renders a `<span>` tag which is controlled by the tag prop. The tag can be overridden with a valid html element or custom tag.
+
+**Props:**
```
pixelbandito, line 7: Minor: We shouldn't have to manually write props out.

We used to get these for free from PropTypes, but since we switched to TS I guess we don't.

I'd support leaving them out and creating an issue to figure out how to generate props documentation from TS.

But if you want to include them manually, that's fine too.

src/components/Typography/Typography.module.css
```diff
@@ -0,0 +1,94 @@
+.h1 {
+  font-size: var(--rvr-font-size-h1);
```
pixelbandito, line 2: Minor: I'd suggest dropping these `.h*` helper classes and the associated `--rvr-font-size-*` variables.

The way this is currently implemented, the `.h1` style will only be accessible by passing it in as a `className` prop. Since we're using CSS modules, that would require consumers to look at the components implementation to get the right classname.

I'd guess that developers are more likely to pick sizes and colors straight from Figma and use the `size` and `color` props.

If you _really_ want shortcuts for things like `h1`, you could do...
```
export const H1 = (props) => <Typography size="xl3" tag="h1" {...props} />
```

Let me know if I'm missing something.

src/components/Typography/Typography.tsx
bkonuwa: Yeah, that works. I'm cleaning up the code to use that instead. Thanks

src/components/Typography/README.md
bkonuwa: Ok, I will remove them.

src/components/Typography/Typography.test.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import Typography from './Typography';
+
+describe('Typography', () => {
+  it('renders', () => {
+    render(<Typography data-testid="Typography-Test" />);
+    expect(screen.getByTestId('Typography-Test')).toBeInTheDocument();
+  });
+
+  it('renders the appropriate tag', () => {
```
pixelbandito, line 13: One more nice to have test (and story)...
You could verify that it renders an arbitrary React component in the `tag` props.

pixelbandito (approved):
🌟 Love it.

mikemfleming (approved):

PR #293, opened 2020-12-16
Form 292

### LINKED ISSUE #292: Form components: logical form
related to #281 #177 #168 #128 

Add a form component that takes care of all state logic.

mikemfleming:
### What's in this pr?
This resolves #292 by adding a Form component that provides state management helpers. It's my first real pr to rover-ui so be brutal!

### ☝ Concerns
What I'm proposing is a home brewed state management system. Would it be better to introduce a lib that's dedicated to the task? 

[Formik](https://github.com/formium/formik) is a really good one IMO but if this is sufficient then we can leave it at that.

### DIFF
--- a/src/components/Form/Form.tsx
+++ b/src/components/Form/Form.tsx
@@ -1,0 +1,114 @@
+import React, { useState, useEffect } from 'react';
+import _isEmpty from 'lodash/isEmpty';
+import _set from 'lodash/set';
+
+const Form = ({
+  children,
+  initialValues = {},
+  validationSchema = {},
+  onSubmit,
+  className,
+}) => {
+  const [formValues, setFormValues] = useState(initialValues);
+  const [touched, setTouched] = useState({});
+  const [validationErrors, setValidationErrors] = useState({});
+  const [isSubmitting, setIsSubmitting] = useState(false);
+  const [formSubmitError, setFormSubmitError] = useState(null);
+
+  const handleBlur = ({ target: { name } }) => {
+    setTouched((prevState) => ({
+      ...prevState,
+      [name]: true,
+    }));
+  };
+
+  const handleSubmit = async (e) => {
+    e.preventDefault();
+
+    if (_isEmpty(validationErrors)) {
+      setIsSubmitting(true);
+    }
+  };
+
+  const handleCustom = (fieldName, callback) => (e) => {
+    e.preventDefault(); // TODO: maybe call this conditionally
+
+    setFormValues((prevFormValues) => ({
+      ...prevFormValues,
+      [fieldName]: callback(prevFormValues[fieldName]),
+    }));
+  };
+
+  const handleChange = ({ target: { name, type, value, checked } }) => {
+    const newValue = type === 'checkbox' ? checked : value;
+
+    setFormValues((prevFormValues) =>
+      _set({ ...prevFormValues }, name, newValue)
+    );
+  };
+
+  useEffect(() => {
+    const validateInputs = () => {
+      const errors = Object.entries(formValues).reduce(
+        (errorsObject, [fieldName, fieldValue]) => {
+          const fieldValidation = validationSchema[fieldName];
+          if (
+            fieldValidation?.validate &&
+            fieldValidation?.message &&
+            !fieldValidation.validate(fieldValue)
+          ) {
+            return {
+              ...errorsObject,
+              [fieldName]: fieldValidation.message,
+            };
+          }
+
+          return errorsObject;
+        },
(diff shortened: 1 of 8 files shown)


src/components/Form/story.tsx
```diff
@@ -0,0 +1,73 @@
...
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+  })
+  .add(
+    'Overview',
+    () => (
+      <Form
+        initialValues={{ nameInput: 'james r. cat' }}
```
pixelbandito, line 17: Just a thought: What should happen if no initial values are provided? Might there be a good way to default to "empty" values?
I guess this could be tough if we don't know up front whether an input will be text, radio, etc.

src/components/Form/story.tsx
```diff
@@ -0,0 +1,73 @@
...
+        // eslint-disable-next-line no-alert
+        onSubmit={(formValues) => alert(JSON.stringify(formValues))}
```
pixelbandito, line 19: This is where, if no submit handler is provided, I think falling back to default form behavior would make sense.

src/components/Form/story.tsx
```diff
@@ -0,0 +1,73 @@
...
+            validate: (value) => value !== '',
+          },
+        }}
+      >
+        {({
+          formState,
+          formValues,
+          handleChange,
+          handleBlur,
+          handleCustom,
```
pixelbandito, line 32: The name `handleCustom` is confusing to me. Is it a change handler, blur handler, submit handler?
Maybe something along the lines of `processFieldValue`?

Would it be better to have it take and update the entire form state? I'm thinking of cases where fields would depend on each other, like if a user chose "February" and we had to reset or modify a date from 31 to 28.

src/components/Form/story.tsx
```diff
@@ -0,0 +1,73 @@
...
+        }) => (
+          <>
+            <label htmlFor="nameInput">
+              Name
+              <br />
+              <input
+                type="text"
+                name="nameInput"
+                value={formValues.nameInput}
```
pixelbandito, line 41: This would be crazy:
Having the user enter `name="foo"` _and_ `formValues.foo` feels redundant.
Would it make sense to iterate the children/dom elements and apply values and event handlers that way? I'm not sure it would be feasible or fast, but the interface would sure be nice if you could just add `<input type="text" name="myInput" />` and everything else was automatic.

We might be able to do something like that with an `EasyInput` component and the React Context API.

pixelbandito:
```<Form initialValues={{
  foo: ""
}}>
<EasyInput name="foo" type="text" />
</Form>
```

src/components/Form/story.tsx
```diff
@@ -0,0 +1,73 @@
...
+        onSubmit={(formValues) => alert(JSON.stringify(formValues))}
+        validationSchema={{
+          nameInput: {
+            message: 'No empty strings',
+            validate: (value) => value !== '',
+          },
+        }}
+      >
+        {({
+          formState,
```
pixelbandito, line 28: If you combined some parts of `formState` and `formValues`, could you do something like:
```
<input
  value={form.email.value}
  onChange={form.email.onChange}
  data-touched={form.email.touched}  
  data-error-messages={form.email.errorMessage}  
/>
```

Or, if you wanted to extract some properties in the function as child argument, you could get as simple as:
```
<Form ... />
  {
    ({ form: { touched, errorMessage, ...email } }) => {
      <input {...email}  />
    }
  }
</Form>
```

pixelbandito:
Arg, I've been meaning to get to this, but it's complex enough I want to run it locally. Will do Monday at the latest. Sorry, @wendigolabs!

src/components/Form/Form.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  const handleBlur = ({ target: { name } }) => {
+    setTouched((prevState) => ({
+      ...prevState,
+      [name]: true,
+    }));
+  };
+
+  const handleSubmit = async (e) => {
+    e.preventDefault();
```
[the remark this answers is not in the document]
pixelbandito: I mean: Should we not prevent the default at all, or just allow the user to prevent default by exposing an `onSubmit` function prop?

One potentially nice thing about `submit` events is that, for short forms, they capture when the user enters text in a field and hitting the "Enter" key, iirc.

I'd hate to trap the submit events and prevent their default behavior if they're useful.

pixelbandito:
Changes so far look good! I just resolved a handful of conversations.

PR #297, opened 2021-01-25
Removed attributes we don't need to explicitly define.

wendigolabs:
The component doesn't do anything special with `readOnly` ~~and `onChange`~~, so there's no reason to privilege these particular attributes in the component.

Also added some more `<input />` types because they're so cool.

### DIFF
--- a/src/components/Input/Input.tsx
+++ b/src/components/Input/Input.tsx
@@ -13,6 +13,7 @@ export interface InputProps
 
 const Input: React.FC<InputProps> = ({
   className = '',
+  onChange = () => {},
   fauxDisabled = false,
   forwardedRef: ref,
   ...passedProps
@@ -23,6 +24,7 @@ const Input: React.FC<InputProps> = ({
         [styles.disabled]: fauxDisabled,
       })}
       ref={ref || undefined}
+      onChange={onChange}
       {...passedProps}
     />
   );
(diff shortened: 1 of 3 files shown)


pixelbandito (approved):

pixelbandito:
> The story file demos these two attributes so since they're no longer specifically referenced in the component, should they be removed? Logically, they should be, since we're not demoing any of the other umpteen `input` attributes that aren't specifically defined in `Input`, but it does enrich the examples to have them...

I hadn't considered that, good observation.
I'd think: go wild with illustrative examples in the "Examples" story, but only instrument used props in the "Overview" story.
That said, having an `input` without `onChange` sounds kinda weird even in the bare-bones overview. I don't mind leaving it in if you don't.

wendigolabs:
> > The story file demos these two attributes so since they're no longer specifically referenced in the component, should they be removed? Logically, they should be, since we're not demoing any of the other umpteen `input` attributes that aren't specifically defined in `Input`, but it does enrich the examples to have them...
> 
> I hadn't considered that, good observation.
> I'd think: go wild with illustrative examples in the "Examples" story, but only instrument used props in the "Overview" story.
> That said, having an `input` without `onChange` sounds kinda weird even in the bare-bones overview. I don't mind leaving it in if you don't.

Agreed, it does make sense to leave that one (`onChange`) in if nothing else.  So I'll do readd that, and add a few more examples.

wendigolabs:
Ok, made the discussed changes.  See how this version looks.

PR #299, opened 2021-02-22
Typography cleanup

pixelbandito:


pixelbandito:
While verifying that `Typography` wasn't being exported in the most recent release, I noticed that it was also missing instrumentation for its props in Storybook.

That led down a short rabbit hole and some additional cleanup.

pixelbandito:
fyi: Typography hasn't been released yet. The code is on master, waiting for a new release and version bump.

bkonuwa (approved):

src/components/Typography/Typography.tsx
```diff
@@ -26,30 +23,32 @@ type TypographyType = React.FC<TypographyProps> & {
   Block: React.FC<TypographyProps>;
 };
 const Typography: TypographyType = ({
-  align = 'inherit',
   color = 'primary',
   className: passedClassName = '',
-  size = 'md',
+  size = 'medium',
```
bkonuwa, line 28: Curious... Why is this defaulted `medium` instead of `md`? The style class is defined as `.size--md` and the size type doesn't have a medium option.

pixelbandito: Oh, I didn't mean to do that! Thank you for catching it!

Issue #305, opened 2021-03-09
Tooltip a11y needs work


pixelbandito:
Misc a11y fixes to bring behavior in line with recommendations for tooltips here: https://knowbility.org/blog/2018/WCAG21-1413ContentHoverFocus 

Known issues:
* Content is rendered in an unintuitive place. 
* Focus / blur doesn't work like hovering.
* In click-to-open mode, the content isn't being read by screen readers.
* The user can't dismiss the tooltip from the keyboard in hover mode.
* The close button is an icon with no text representation.

pixelbandito:
Fixed in [v2.6.0](https://github.com/cision/rover-ui/releases/tag/v2.6.0)

Issue #298, opened 2021-02-09
More alignment options for Tooltip


pixelbandito:
Currently, the tooltip supports left, right, top, and bottom.

A consumer is asking about the ability to add right-top alignment.

![image](https://user-images.githubusercontent.com/304177/107419059-f9ba0800-6adc-11eb-942c-17fc5239d5e8.png)

Things to consider:

_There are 12 alignments possible._
 
E.g.
- "right-top": The tooltip is to the right of the trigger, justified top.
- "top-right": The tooltip is above the trigger, justified to the right.

So for full position support, we'd have
- top
- top-right
- top-left
- right
- right-top
- right-bottom
- ...etc.

_The amount of the top / bottom offset in this example could be tricky to calculate_

Should the height / width of the tooltip and trigger be into account in case the tooltip is smaller than the trigger? 
Are we comfortable with a more naive approach like "top align, then offset the arrow by a standard padding + half of a standard line height"?

pixelbandito:
Deployed in [v2.6.0](https://github.com/cision/rover-ui/releases/tag/v2.6.0)

PR #307, opened 2021-03-18
Tooltip alignment options, a11y improvements

pixelbandito:
Addresses issues #305 and #298 

* Adds an optional secondary dimension to Tooltip config, e.g. `top-left`.
* Uses `CSSTransition` to unmount tooltips when they're closed and clean up fade in / fade out.
* a11y improvements
  * "hover" tooltips are also controlled with focus/blur
  * users can dismiss tooltips (even hover/focus ones) with the Escape key
  * you can mouse from a tooltip trigger to the tooltip content in hover mode without it disappearing


package.json
```diff
@@ -48,8 +48,8 @@
     "pre-commit:lint": "yarn lint-staged",
     "pre-commit:types": "yarn tsc",
     "pre-commit:test": "yarn test",
-    "test": "cross-env CI=1 react-scripts test --env=jsdom",
-    "test:watch": "react-scripts test --env=jsdom",
+    "test": "cross-env CI=1 react-scripts test --env=jest-environment-jsdom-sixteen",
```
pixelbandito, line 51: This was necessary for some tests to work with hover / focus.

package.json
```diff
@@ -88,10 +88,10 @@
     "@storybook/addons": "^5.3.19",
     "@storybook/preset-create-react-app": "^3.0.0",
     "@storybook/react": "^5.3.19",
-    "@testing-library/jest-dom": "^5.10.1",
-    "@testing-library/react": "^10.2.1",
-    "@testing-library/react-hooks": "^3.3.0",
-    "@testing-library/user-event": "^12.0.7",
+    "@testing-library/jest-dom": "5.11.9",
```
pixelbandito, line 91: This stuff was just cleanup, trying to fix an issue that ended up being unrelated.

src/components/Tooltip/story.tsx
```diff
@@ -62,74 +54,62 @@ const Element: React.FC<React.HTMLAttributes<HTMLDivElement>> = (props) => (
   <div className="mr-5" {...props} />
 );
 
-const Overview = () => {
```
pixelbandito, line 65: When this component was defined outside the story's render function, knobs didn't work.

mikemfleming (approved):

package.json
pixelbandito: Nerp, after I fixed some other stuff, I didn't need jsdom 16 after all. And including it broke backwards compatibility with Node 10 (which we say we support).
So I removed it, but I'm kinda sad that Node 10 support might stop us from upgrading stuff, in general.

bkonuwa (approved):

PR #309, opened 2021-03-29
feat: Add Modal component

bkonuwa:


src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+interface ModalProps {
+  children?: React.ReactNode;
+  className?: ClassValue;
+  isOpen?: boolean;
+  onClose?: () => void;
+  size?: Size;
+  style?: CSSProperties;
+}
+
+type ModalChildProps = Pick<ModalProps, 'children' | 'className'>;
```
pixelbandito, line 21: Just a thought: I wonder if we should add a default component interface that includes `className` and `children`.
Almost all of our components do / should support those.

At least, we could add it to the hygen template.
We could also add a basic `Component.tsx` file to the src root that exports it so the basic types can be extended easily.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+      }}
+    >
+      <div
+        role="presentation"
+        {...passedProps}
+        style={passedStyle}
+        className={styles.modal}
+        onClick={onClose}
+      >
+        <div
```
pixelbandito, line 105: From that same a11y article, I think the modal should have the role "dialog". Not sure if it should be here or on this div's parent, though.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+        enterDone: styles.enterDone,
+        exit: styles.exit,
+      }}
+    >
+      <div
+        role="presentation"
+        {...passedProps}
+        style={passedStyle}
+        className={styles.modal}
+        onClick={onClose}
```
pixelbandito, line 103: Looking at an article about a11y in modals, I like the idea of creating a modal context and making the `onClose` function available to children that way.
https://medium.com/@seif_ghezala/how-to-create-an-accessible-react-modal-5b87e6a27503

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+      >
+        <div
+          role="presentation"
+          className={modalContentStyle}
+          onClick={(e) => e.stopPropagation()}
```
pixelbandito, line 108: There's always a chance that a consumer will have a good reason for wanting to capture clicks inside the modal, from a parent.

Instead of using `stopPropagation`, you could handle this use case with a `handleClickBackground` function.
That function would look at the click event's target, and only call `onClose` if the event target is not inside the modal.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+  children,
+  className: passedClassName = '',
+  isOpen = false,
+  onClose = () => {},
+  size = 'md',
+  style: passedStyle = {},
+  ...passedProps
+}) => {
+  const handleEscape = useCallback(
+    (e: KeyboardEvent) => {
```
pixelbandito, line 42: You should probably check to see if the cursor is focused on an element inside the modal before you call `onClose`

For instance, in your storybook implementation:
* Open the large modal (with the form)
* Start typing something
* Hit Escape
* I expect that the current input loses focus
* But actually, the modal closes (technically the input loses focus as well)

I think you can do this by getting the `document.activeElement` and a ref for your modal, then add a condition like
```
if (modalRef.current && !modalRef.current.contains(document.activeElement)) {
  onClose();
}
```

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+    if (!isOpen) {
+      window.removeEventListener('keyup', handleEscape);
+    }
+
+    return () => {
+      if (isOpen) window.removeEventListener('keyup', handleEscape);
+    };
+  }, [handleEscape, isOpen]);
+
+  useEffect(() => {
```
pixelbandito, line 64: This is probably good for a follow-up PR:

It would be nice to autofocus inside the modal when it opens, and add a focus trap.

Here's the general idea of the focus trap: https://medium.com/@im_rahul/focus-trapping-looping-b3ee658e5177
And here's an implementation of getting all focusable elements inside a div or the document: https://zellwk.com/blog/keyboard-focusable-elements/

There's a catch mentioned about focus trapping with screen readers, and I don't have a solution for that, but just implementing the auto-focus would still be really useful.

src/shared/variables.css
```diff
@@ -25,4 +25,5 @@
 
   /* Borders */
   --rvr-border-radius: 4px;
+  --rvr-tile-border-radius: 6px;
```
pixelbandito, line 28: This is nitpicky, but I'd rename this one, twice. Here's why:
For any css vars with multiple values (like border-radius, now), the convention should be
* Have an ordered list of the values that doesn't use any semantic names. Think of this like a color palette.
* Then add a semantic section that maps those values to their use cases.

In this case, it would be something like:
```
/* Palette */
--rvr-border-radius-md: 4px;
--rvr-border-radius-lg: 6px;

/* Semantic-ish */
--rvr-border-radius-tile: var(--rvr-border-radius-md);
```

That lets us change the whole feel of the app by changing the "palette" values, but it also gives developers a cheat sheet to know what size we typically use for tile-type components.

pixelbandito, line 28: Separate naming thing:
I think I would label this border radius with "dialog" instead of "tile"

There's no tight definition, but I found the breakdown on this page helpful. https://ux.stackexchange.com/questions/78798/whats-the-difference-between-cards-panels-and-tiles

The article suggests that "tile" is mostly associated with those square in Microsoft's Metro UI, and those aren't like modals at all.

I don't think modals quite fit into any of those categories, but I can imagine we'd end up with a few types of dialog box that should share the same border radius.

pixelbandito (commented):
Sorry, lot of words.
I checked out the code and overall, I'm really happy with how it looks and functions!

So please bear with my nitpicks, and feel free to push back on any of them.

src/components/Modal/Modal.tsx
```diff
@@ -0,0 +1,160 @@
...
+      classNames(
+        styles.modalContent,
+        {
+          [styles[`size--${size}`]]: size,
+        },
+        passedClassName
+      ),
+    [passedClassName, size]
+  );
+  return ReactDOM.createPortal(
```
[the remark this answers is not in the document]
bkonuwa: Yeah, I'm not sure there's a catch. That's how portal is used in this react TS cheet sheet. https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/portals/

bkonuwa:
> Sorry, lot of words.
> I checked out the code and overall, I'm really happy with how it looks and functions!
> 
> So please bear with my nitpicks, and feel free to push back on any of them.

No worries. Thanks for the feedback. I will add these suggestions this week.

pixelbandito (approved):
🌹 🍔

src/components/Modal/Modal.test.tsx
```diff
@@ -0,0 +1,83 @@
...
+  });
+
+  describe('onClose callback', () => {
+    it('calls onClose callback when the escape key is pressed', async () => {
+      const onClose = jest.fn();
+      render(<Modal isOpen onClose={onClose} data-testid="Modal-Test" />);
+      const modal = screen.getByTestId('Modal-Test');
+      expect(modal).toBeInTheDocument();
+
+      fireEvent.keyDown(modal, {
```
dkordik, line 75: `userEvent.type` would probably be preferred here to simulate an actual user escape press
https://testing-library.com/docs/ecosystem-user-event/#typeelement-text-options

dkordik (changes requested):

dkordik (approved):

PR #311, opened 2021-05-03
Feature ever 13205 a11y buttons and links

pixelbandito:
**Summary**
Changes button, link, and most other colors that formerly used `#0092c2` to use `#007fa9`, which passes color contrast checks when used on a white background (or as a background behind white text).

This should make our primary buttons and link-style faux-buttons pass a11y contrast checks.

**Refactoring**
I also replaced some component references to the palette's `--rvr-blue` variable with the semantically-appropriate `--rvr-color-primary` and `--rvr-color-info` variables.

Palette colors aren't ever intended for direct usage in components. Within components, everything should be mapped to a meaningful semantic color variable.

**Exceptions**
The one thing I overwrote was `input:focus`'s border and glow colors. Since the input focus state is transient and not related to text legibility, I mapped that to our `--rvr-primary-hover` color instead. The glow looked weird with the dark color.

So focused inputs still look like this: 
image

### DIFF
--- a/example/src/App.js
+++ b/example/src/App.js
@@ -26,6 +26,7 @@ import {
   Input,
   InputTime,
   Typography,
+  Modal,
   // IMPORT_INJECTOR
 } from '@cision/rover-ui';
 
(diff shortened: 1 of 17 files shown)


bkonuwa (approved):

pixelbandito:
Hrm, not sure why this PR shows @bkonuwa's modal changes - I already synced v3-alpha branch with master.

Gonna wait an hour or so and see if Github picks up the updates to this PR - it should only contain 1 commit.

PR #317, opened 2021-05-24
feat: EVER-13206: a11y focus states for buttons and links

pixelbandito:


dkordik (approved):

src/components/Button/story.tsx
```diff
@@ -179,43 +181,75 @@ storiesOf('Planets/Button', module)
...
+                    (stateCombo) => {
+                      const key = stateCombo.join('-');
+
+                      const trueStates = stateCombo.reduce(
+                        (result, state) => ({
+                          ...result,
+                          [state]: true,
+                        }),
+                        {}
+                      );
```
dkordik, line 238: another way to do this might be something like this:
```javascript
const trueStates = Object.fromEntries(
    stateCombo.map((state) => [state, true])
  );
```
not sure why, but I always need to stare at `.reduce`'s for a while to grok them. but your way feels very React-y

bkonuwa (approved):

src/components/Button/story.tsx
pixelbandito: You're right, it's really hard to read. I'll take a crack at refactoring. I didn't even know about `fromEntries`, I have a feeling we're going to be friends.

PR #324, opened 2021-07-19
Feature/rover UI/ever 13326/kite component

bkonuwa:
feat: Add basic kite component
image

### DIFF
--- a/src/components/Kite/Kite.module.css
+++ b/src/components/Kite/Kite.module.css
@@ -1,0 +1,71 @@
+.Kite {
+  z-index: var(--rvr-zindex-kite);
+  visibility: hidden;
+  box-sizing: border-box;
+  position: fixed;
+  top: 12px;
+  right: 12px;
+  transition: transform .6s ease-in-out;
+  animation: slide-in-right .7s;
+  display: flex;
+  flex-flow: column nowrap;
+  align-items: center;
+  justify-content: center;
+  background-color: var(--rvr-white);
+  width: 342px;
+  height: 56px;
+  box-shadow: 0px 5px 15px rgba(0, 0, 0, 0.2);
+  border-radius: 4px;
+}
+
+.visible {
+  visibility: visible;
+}
+
+.content {
+  align-items: center;
+  display: flex;
+  width: 100%;
+}
+
+.icon {
+  margin-left: 15px;
+}
+
+.title {
+  flex-grow: 2;
+  font-size: var(--rvr-font-size-md);
+  color: var(--rvr-color-font-dark);
+  margin-left: 15px;
+}
+
+.enterDone {
+  opacity: 1;
+  pointer-events: visible;
+  transform: translateY(0);
+}
+
+.exit {
+  opacity: 0;
+  transform: translateY(-200px);
+}
+
+.top-right {
+  top: 12px;
+  right: 12px;
+  transition: all .6s ease-in-out;
+  transform: translateX(100%);
+}
+
+
+
+
+@keyframes slide-in-right {
+  from {
+    transform: translateX(100%);
+  }
+
+  to {
+    transform: translateX(0);
+  }
+}
--- a/src/components/Kite/Kite.module.css
+++ b/src/components/Kite/Kite.module.css
@@ -28,7 +28,6 @@
   display: flex;
   flex-direction: row;
   width: 100%;
-  justify-content: space-between;
   align-items: center;
 }
 
@@ -37,7 +36,11 @@
   min-width: 0;
   display: flex;
   align-items: center;
-  width: 80%;
+  width: 100%;
+}
+
+.title {
+  flex-grow: 2;
 }
 
 .dismissButton {
@@ -53,7 +56,6 @@
 
 .enterDone {
   opacity: 1;
-  pointer-events: visible;
   transform: translateX(0);
 }
 
--- a/src/components/Kite/Kite.module.css
+++ b/src/components/Kite/Kite.module.css
@@ -8,14 +8,17 @@
   transition: transform .6s ease-in-out;
   animation: slide-in-right .7s;
   display: flex;
-  flex-flow: column nowrap;
+  flex-flow: row nowrap;
   align-items: center;
   justify-content: center;
   background-color: var(--rvr-white);
   width: 342px;
-  height: 56px;
+  /*height: 56px;*/
   box-shadow: 0px 5px 15px rgba(0, 0, 0, 0.2);
   border-radius: 4px;
+  padding: 16px;
+  font-size: var(--rvr-font-size-md);
+  color: var(--rvr-color-font-dark);
 }
 
 .visible {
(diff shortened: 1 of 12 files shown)


pixelbandito:
> feat: Add basic kite component
> image

I think the the text should be left-aligned between the icon and close button. https://www.figma.com/file/19I1wJEAQtTInpL53dLHNx/Rover-UI?node-id=386%3A92

src/components/Kite/Kite.module.css
```diff
@@ -0,0 +1,88 @@
+.Kite {
+  z-index: var(--rvr-zindex-kite);
+  visibility: hidden;
+  box-sizing: border-box;
+  position: fixed;
+  top: 12px;
```
pixelbandito, line 6: shared/sizing.css probably has a variable you can use here.

src/components/Kite/Kite.module.css
```diff
@@ -0,0 +1,88 @@
...
+  right: 12px;
+  transition: transform .6s ease-in-out;
```
pixelbandito, line 8: _No action needed for now_, but we should ask the design team about defining a set of standard transitions and box shadows.

My gut instinct is that Kites should probably use the same shadows as modals. They're all things that have similar perceived "altitudes" and visual weights, maybe?

Seems like our CSS variables could have something like a low-altitude / low contrast box shadow for tooltips and dropdown menus, and a high-altitude / high contrast box shadow for elements that float "further away" from the page content.

For transitions, we'd need to think a bit more. There's a heuristic for moving smaller items more slowly than larger items, and it would probably depend on whether the item is moving or just fading in/out. Regardless, it would be nice to have a formula and a few useful variables so we're consistent in our approach across components.

src/components/Kite/Kite.module.css
```diff
@@ -0,0 +1,88 @@
...
+.content {
+  flex: 1 1 auto;
+  min-width: 0;
+  display: flex;
+  align-items: center;
+  width: 80%;
+}
+
+.childrenWrapper {
+  margin-top: 8px;
```
pixelbandito, line 44: Same thing, these should use variables.

src/components/Kite/Kite.module.css
```diff
@@ -0,0 +1,88 @@
...
+  justify-content: space-between;
+  align-items: center;
+}
+
+.content {
+  flex: 1 1 auto;
+  min-width: 0;
+  display: flex;
+  align-items: center;
+  width: 80%;
```
pixelbandito, line 40: Rather than width: 80%, I think we should set the content to use all the remaining space once the close button, icon, and padding / margins have been accounted for.

src/components/Kite/Kite.module.css
```diff
@@ -0,0 +1,88 @@
...
+
+.icon {
+  display: flex;
+  flex: 0 0 auto;
+  margin-right: 16px;
+}
+
+.enterDone {
+  opacity: 1;
+  pointer-events: visible;
```
pixelbandito, line 61: I wasn't familiar with `visible` as an option for `pointer-events`, but it looks like it's for SVG only.
In this case, you probably don't even need it, since you're not setting `pointer-events` to a value other than `auto` in the first place.

src/components/Kite/Kite.module.css
```diff
@@ -0,0 +1,88 @@
...
+
+@keyframes slide-in-right {
+  from {
+    transform: translateX(100%);
+  }
+
+  to {
+    transform: translateX(0);
+  }
+}
```
pixelbandito, line 88: We should have newlines at the end of files - I think this symbol means it's missing.
Most editors have an option to add a newline automatically on save.
https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

src/components/Kite/Kite.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+import ReactDOM from 'react-dom';
+import { CSSTransition } from 'react-transition-group';
+import styles from './Kite.module.css';
+import Button from '../Button';
+import Icon from '../Icon';
+
+interface KiteProps {
+  children?: React.ReactNode;
+  className?: ClassValue;
```
pixelbandito, line 13: Imo, we shouldn't use `ClassValue`.
I'm removing it in an upcoming PR, hopefully.

Our `className` type should just be an optional string.

(Using `ClassValue` would allow users to pass anything that the `classnames` package can handle, e.g. arrays, objects with optional extra class names. That's fine, and might be nice to have, but 1) we don't document it, 2) it causes some TS headaches, and 3) I think consumers can just use the `classnames` package on their own and send us full composed strings.

src/components/Kite/story.tsx
```diff
@@ -0,0 +1,173 @@
...
+storiesOf('Galaxies/Kite', module)
+  .addParameters({
+    readme: {
+      sidebar: Readme,
+    },
+  })
+  .add(
+    'Overview',
+    () => (
+      <Kite visible={false} ttl={3000} onClose={() => {}}>
```
pixelbandito, line 133: I'd use storybook knobs to control the `visible` and `ttl` props, and a storybook action to give the user feedback when `onClose` is called.

src/components/Kite/story.tsx
```diff
@@ -0,0 +1,173 @@
...
+        <Kite.Header canBeDismissed title="Kite Title" onClose={() => {}}>
```
pixelbandito, line 134: Just curious: Does `Kite.Header` need its own `onClose` function? Wouldn't it be redundant to the function on `Kite`?

src/components/Kite/story.tsx
```diff
@@ -0,0 +1,173 @@
...
+          <Kite.Icon
+            fill="green"
+            height="20"
+            name="check"
+            style={{ display: 'block' }}
```
pixelbandito, line 139: What do you think of making the icon size and `display: block` part of the component's built-in styles?
(Or are there already defaults, and this is showing the user how to make overrides?)

pixelbandito (changes requested):

src/components/Kite/story.tsx
bkonuwa: Yeah, the `Kite.Header` need it's own `onClose` function because the `canBeDismissed` prop is set in `Kite.Header`. If it's set to true, we need the close button to call the onClose handler. I couldn't  figure out a way to pass the `onClose` function from the `Kite` to `Kite.Header`. If there's a better way to accomplish this, I'm all ears.

src/components/Kite/story.tsx
bkonuwa: I'm using the props that are already defined on Rover Icon component. There are defaults and this is showing how to make overrides.

src/components/Kite/story.tsx
pixelbandito: I've done this with Context and by iterating over the children in the parent component and manually spreading props.
The context way is cleaner.

It might be better to do this sooner rather than later, so we don't end up with a bunch of `Kite.Header` components that have extra props. I think we can put both `canBeDismissed` and `onClose` as top-level `Kite` props.

Also, does the title need to be a string? If it's just getting rendered in the header, I think we should pass the header content as children and allow any React node as the type.

pixelbandito (approved):

src/components/Kite/story.tsx
pixelbandito: For future readers:
Ended up moving `canBeDismissed` and `onClose` props to the `Kite` component, and nixing the context idea.

Issue #291, opened 2020-12-16
Buttons should be capable of ref forwarding


pixelbandito:
We need this to migrate some current usage.

We should decide if ref forwarding will be automatic going forward, or if we want to provide it as a named component variant.

sieglerz:
Tracking with EVER-12930

PR #325, opened 2021-07-19
Feature: EVER-13451: Add `Select` component

pixelbandito:


_templates/component/new/Component.tsx.t
```diff
@@ -5,22 +5,19 @@ unless_exists: true
 <% PascalName = h.changeCase.pascal(name) %>
 import React from 'react';
 import classNames from 'classnames';
-import { ClassValue } from 'classnames/types';
 
 import styles from './<%= PascalName %>.module.css';
 
 interface <%= PascalName %>Props {
-  className?: ClassValue;
+  className?: string;
```
pixelbandito, line 12: I wanted to do this for a while, but my lazy justification for doing it now is that `ClassValue` is no longer available from the given import statement. I'm sure I could've found an alternative import, but it didn't seem worth it.

We were never intentionally exposing all the config options of `classNames` to consumers. They can pass plain old string class names with spaces between them, and that should be enough.

This also cleans up a handful of TS edge cases.

src/components/Button/Button.tsx
```diff
@@ -20,10 +20,12 @@ interface BaseButtonProps {
   circle?: boolean;
   className?: string;
   darkMode?: boolean;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  forwardedRef?: React.Ref<any>;
```
pixelbandito, line 24: I need to go back to TS school. This component can spit out Refs to at least 3 element types, and I've had a helluva time getting them to play nicely when composing components.

If anyone feels really confident in this area, I'd welcome a helping hand.

src/components/Dropdown/Menu/Item/Item.module.css
```diff
@@ -1,36 +1,15 @@
 .Item {
-  background: transparent none;
```
pixelbandito, line 2: I'm now defaulting `Dropdown.Menu.Item` to use `Button` components with `level="text"`, which eliminates the need for a bunch of this CSS.

package.json
```diff
@@ -48,9 +48,9 @@
     "pre-commit:lint": "yarn lint-staged",
     "pre-commit:types": "yarn tsc",
     "pre-commit:test": "yarn test",
-    "test": "cross-env CI=1 react-scripts test --env=jsdom",
-    "test:watch": "react-scripts test --env=jsdom",
-    "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache",
+    "test": "cross-env CI=1 react-scripts test --env jest-environment-jsdom-fourteen",
+    "test:watch": "react-scripts test --env jest-environment-jsdom-fourteen",
+    "test:debug": "react-scripts --inspect-brk test --runInBand --no-cache --env jest-environment-jsdom-fourteen",
```
pixelbandito, line 53: When I tried to use `waitFor`, our current tests failed with "TypeError: MutationObserver is not a constructor #731".

I found [this issue documented](https://github.com/testing-library/react-testing-library/issues/731), that suggests using a different jest environment jsdom version.

People in that thread recommended both fourteen and sixteen, and link to a doc recommending fifteen.
sixteen works, so I'm going with that.

src/components/Button/Button.tsx
```diff
@@ -35,19 +37,21 @@ interface BaseButtonProps {
 }
 
 // Button props
-type ButtonElementProps = React.ButtonHTMLAttributes<HTMLButtonElement> &
+export type ButtonElementProps = React.ButtonHTMLAttributes<HTMLButtonElement> &
```
pixelbandito, line 40: I switched `Dropdown.Menu.Item` to use `Button` components, and exported this for reuse over there.

src/components/Button/Button.tsx
```diff
@@ -170,4 +176,8 @@ const Button: ButtonType = ({
 
 Button.Addon = Addon;
 
+Button.WithRef = React.forwardRef((props, ref) => (
```
pixelbandito, line 179: This addresses issue #291, in addition to helping me trigger focus on the current selected option when a `Select` is opened.

package.json
```diff
@@ -104,7 +104,7 @@
     "@typescript-eslint/parser": "^2.26.0",
     "auto-changelog": "^2.0.0",
     "babel-eslint": "^10.1.0",
-    "babel-loader": "^8.1.0",
+    "babel-loader": "8.1.0",
```
pixelbandito, line 107: Dumb fix for this issue that popped up when running `yarn test` after installing some new dependencies:
```There might be a problem with the project dependency tree.
It is likely not a bug in Create React App, but something you need to fix locally.

The react-scripts package provided by Create React App requires a dependency:

  "babel-loader": "8.1.0"

Don't try to install it manually: your package manager does it automatically.
However, a different version of babel-loader was detected higher up in the tree:

  /Users/chrisgarcia/Code/Cision/rover-ui/node_modules/babel-loader (version: 8.2.2)

Manually installing incompatible versions is known to cause hard-to-debug issues.
```

src/components/Pill/Pill.tsx
```diff
@@ -1,15 +1,14 @@
...
-import { ClassValue } from 'classnames/types';
 import Addon from './Addon';
 
 import styles from './Pill.module.css';
 
 export interface PillProps
   extends Omit<React.HTMLAttributes<HTMLDivElement>, 'className' | 'onClick'> {
   checked?: boolean;
-  className?: ClassValue;
+  className?: string;
```
pixelbandito, line 11: This line (and the Omit counterpart) can be omitted since className now matches the type already on `div`

PR #336, opened 2021-08-24
Chore: Deprecate Node 10.x support

pixelbandito:
This will semantically require a major version, since dropping support for Node 10.x is a breaking change.
However, an internal poll of known consumers indicates _no one_ is using Node 10, and it's no longer a long-term support (LTS) version; so the impact should be minimal to non-existent.

### DIFF
--- a/_templates/component/new/Component.tsx.t
+++ b/_templates/component/new/Component.tsx.t
@@ -5,22 +5,19 @@ unless_exists: true
 <% PascalName = h.changeCase.pascal(name) %>
 import React from 'react';
 import classNames from 'classnames';
-import { ClassValue } from 'classnames/types';
 
 import styles from './<%= PascalName %>.module.css';
 
 interface <%= PascalName %>Props {
-  className?: ClassValue;
+  className?: string;
 }
 
 const <%= PascalName %>: React.FC<<%= PascalName %>Props> = ({
-  className: klassName = '',
+  className = '',
   ...props
 }) => {
-  const className = classNames(styles.<%= PascalName %>, klassName);
-
   return (
-    <div {...props} className={className}>
+    <div {...props} className={classNames(styles.<%= PascalName %>, className)}>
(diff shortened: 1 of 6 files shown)


_templates/component/new/Component.tsx.t
```diff
@@ -5,22 +5,19 @@ unless_exists: true
 <% PascalName = h.changeCase.pascal(name) %>
 import React from 'react';
 import classNames from 'classnames';
-import { ClassValue } from 'classnames/types';
```
pixelbandito, line 8: 🙄 This is something like the 3rd branch in which I've pushed this change. Hopefully this is the first one to merge to master.

package.json
```diff
@@ -15,7 +15,7 @@
   "module": "dist/index.module.js",
   "jsnext:main": "dist/index.es.js",
   "engines": {
-    "node": ">=10"
```
pixelbandito, line 18: The big thing! node 10 isn't even in maintenance LTS anymore.
https://nodejs.org/en/about/releases/

src/stories/storybook-helpers.tsx
```diff
@@ -66,9 +69,7 @@ export const InteractiveInput: React.FC<InteractiveInputProps> = ({
   defaultValue,
   ...passedProps
 }) => {
-  const [value, setValue] = useState<string | string[] | number | undefined>(
-    defaultValue
-  );
+  const [value, setValue] = useState<InputValue>(defaultValue);
```
pixelbandito, line 72: Was throwing a type error because `defaultValue` was being seen as readonly when it wasn't explicitly typed.

wendigolabs: That's funny, I was running into typing issues on `Form`, related to various typed inputs, and I think my hack to get it to pass was no bueno.  I'll try to steal this...

wendigolabs (approved):
🌮

PR #341, opened 2021-09-01
Ever 15000 rover UI add toggle switch component

lihnick:


example/src/App.js
```diff
@@ -415,6 +417,20 @@ const App = () => {
         {JSON.stringify(inputCheckboxValue)}
       </Section>
 
+      <Section title="Input.Toggle">
+        <h1>Input.Toggle</h1>
+        <Toggle
+          onClick={(e) =>
+            console.log(e.target.checked) ||
```
pixelbandito, line 424: Huh, I just saw Namanda leave this too. I'm guessing y'all both copy/pasted from my `Checkbox` component example.
It won't hurt anything, but you can remove it if you like, and remove the one from checkbox as well.

src/components/Input/Toggle/Toggle.test.tsx
```diff
@@ -0,0 +1,24 @@
...
+import '@testing-library/jest-dom';
+
+import Toggle from './Toggle';
+
+describe('Toggle', () => {
+  it('renders', () => {
+    render(<Toggle readOnly />);
+  });
+
+  describe('when rendered with id attribute', () => {
```
pixelbandito, line 13: Minor, but a test case that it calls the `onChange` handler with the expected value would be nice.
I believe the expected value would be an event with a target element that has the correct `checked` attribute.

Could easily be a follow-on ticket, if you prefer.

pixelbandito (approved):
Minor suggestions, I don't think they would block merging.
👍

pixelbandito (approved):

bkonuwa (approved):

PR #352, opened 2021-11-05
Update CONTRIBUTING to call out presentational components

dkordik:
This could use more cleanup, but just trying to get Jira Github integration working with this particular update
UF-86 #comment does this work?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,10 @@
 # Contributing to RoverUI
 
-## What to avoid in RoverUI
+## Do
+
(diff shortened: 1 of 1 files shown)


dkordik:
might leave this open for a while until we resolve Jira connectivity. using this as our test commits/branch/PR

dkordik:
We got the Jira-GitHub wireup working. Looks like there is no way to surface Jira info in GitHub, so maybe we'll just need to start posting links to our tickets, a la https://cision.atlassian.net/browse/UF-86

lihnick (approved):

